PR #342, opened 2019-07-08
YouTube videos support

### LINKED ISSUE #282: Update gallery component to a more extensible one
Update to use https://github.com/MurhafSousli/ngx-gallery which can allow embedding youtube videos.

### LINKED ISSUE #283: Allow mappers to link a youtube video to their map description
To allow the maps to have an "overview" video if need be, similar to how the Steam store page has trailers for games. Requires #282 to be able to display it.

### LINKED ISSUE #341: Allow embedding YouTube videos for maps
In accordance with #282 , we need to be able to support youtube videos in the map info table to be able to actually show them.

[youtubesurfvideos JSON data from BorkChops](https://github.com/momentum-mod/website/files/3364441/youtubesurfvideos.txt)

MantasAD:
Allows YouTube video to be specified when uploading map.
Allows YouTube video to be displayed in map page.
Fixes map file downloading.
Fixes loading screen disappearing before first content is rendered.
Fixes white color appearing while showing search toggle animation.

Note: gallery configs might be adjusted in the future for improvements.

Closes #282. Closes #283. Closes #341.


client/src/app/pages/dashboard/maps/map.module.ts
```diff
@@ -25,6 +27,7 @@ import {MapListItemComponent} from './map-list/map-list-item/map-list-item.compo
 import {MapEditComponent} from './map-edit/map-edit.component';
 import { MapFavoritesComponent } from './map-favorites/map-favorites.component';
 import { MapListComponent } from './map-list/map-list.component';
+import 'hammerjs';
```
Gocnak: What is this used for?

client/src/app/pages/dashboard/maps/upload-form/map-upload-form.component.ts
```diff
@@ -25,6 +25,8 @@ export interface ImageFilePreview {
   file: File;
 }
 
+var youtubeRegex = /[a-zA-Z0-9_-]{11}/;
```
Gocnak:
```suggestion
const youtubeRegex = /[a-zA-Z0-9_-]{11}/;
```

`var` should never be used.

client/src/app/pages/dashboard/maps/map.module.ts
Gocnak: Further investigation shows that the new gallery component needed it, but they recommend doing this import in `src/polyfills.js`

https://murhafsousli.github.io/ngx-gallery/#/getting-started/core

MantasAD: I put it here only because of the fact that if preloadingStrategy would not be set to PreloadAllModules then angular would not load this script on every page refresh.

MantasAD: Also for some reasons hammerjs is using up to 100% CPU just by moving the mouse around resulting in javascript event loop blocking and on mobile next button presses is sometimes not registered. Maybe other parts of the website are causing this because it's not the case in the example page. Hammerjs only powers optional gestures feature so I think it could be removed.

client/src/polyfills.ts
```diff
@@ -63,6 +63,7 @@ import 'zone.js/dist/zone';  // Included with Angular CLI.
 /***************************************************************************************************
  * APPLICATION IMPORTS
  */
+import 'hammerjs';
```
Gocnak: If the CPU problem still exists should we just remove this? Are the gesture controls worth it?

MantasAD: I was concerned about it at first but then I build website for production and everything was very smooth. So it appears to be a development environment issue only and we can keep this.

Gocnak: Sounds good, if it ever is a problem locally we'll yoink it.

PR #346, opened 2019-08-09
Update user follow notification settings modal

### LINKED ISSUE #313: Use proper (not-enum) strings for notification types
image
Should be like 'Notify when <user> achieves a PB' instead of 'PB_ACHIEVED'

Corey28:
Update content and small styling changes
Closes #313


client/src/app/pages/dashboard/profile/profile-follow/profile-follow.component.ts
```diff
@@ -12,29 +12,49 @@ import {UserFollowObject} from '../../../../@core/models/follow.model';
...
       </nb-card-footer>
     </nb-card>
   `,
-  styles: [`button { margin: 1rem; }`],
+  styles: [`
+    button { margin: 1rem; }
+    nb-card { min-width: 300px; }
+  `],
 })
 export class ProfileNotifyEditComponent implements OnInit {
```
Gocnak: We should probably move this component to its own class, Angular code guidelines typically suggest one class per file.

This can include making the HTML and SCSS files for it, too.

Issue #308, opened 2019-04-16
Merge server/models/db and server/models using new class-based Sequelize models


Gocnak:
As found in [model definition](http://docs.sequelizejs.com/manual/models-definition.html), models are supposed to be classes that extend the Sequelize Model class, and call `<ModelName>.init({}. {sequelize});` to actually init.

This means the models we have outside of the db folder can merge with those inside, as those methods can go inside of the model class itself, and all we do is export the model's class, instead of all the functions.

TheRealHona:
Added `Rewrite: Initial` tag since this would become redundant

PR #349, opened 2019-08-23
Fix avatar ban not being applied to user avatar image

Corey28:
Note that the TODO added in this PR will most likely be removed when working on issue #308

### DIFF
--- a/server/src/models/db/user.js
+++ b/server/src/models/db/user.js
@@ -1,4 +1,5 @@
 'use strict';
+const config = require('../../../config/config');
 
 module.exports = (sequelize, type) => {
 	return sequelize.define('user', {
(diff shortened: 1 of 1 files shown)


server/src/models/db/user.js
```diff
@@ -24,8 +25,14 @@ module.exports = (sequelize, type) => {
 		avatarURL: {
 			type: type.VIRTUAL,
 			get() {
-				const val = this.getDataValue('avatar');
-				return val ? `https://steamcdn-a.akamaihd.net/steamcommunity/public/images/avatars/${val}` : null;
+				const bans = this.getDataValue('bans');
+				const isAvatarBanned = bans & 1 << 2; // TODO: Refactor however needed to use Ban 'enum' (cyclic dep issue occurs when requiring user model in this file)
+				if (isAvatarBanned) {
+                    return config.baseUrl + '/assets/images/blank_avatar.jpg';
```
Gocnak: Spaces mixed with tabs?

PR #350, opened 2019-08-31
Move DB force sync from server start to a separate script

Corey28:


server/config/sqlize.js
```diff
@@ -165,14 +165,8 @@ UserMapRank.belongsTo(User, { foreignKey: 'userID'});
...
 if (env !== 'test') {
-	forceSyncDB().then(() => {
-		// Create our default XP systems table if we don't already have it
-		return xpSystems.initXPSystems(XPSystems);
-	}).then(() => {
-		console.log(`Database & tables created!`)
-	}).catch(err => {
-		console.error(err);
-	});
+    // Create our default XP systems table if we don't already have it
```
Gocnak: Mix of spaces and tabs again.

Perhaps we should look into getting a linter, soon.

Gocnak: You can edit this and force-push it to be right. No need to pollute the history over this.

Corey28: Dang haha x) Yeah, a linter would probably be nice. Will do a force-push this time

PR #351, opened 2019-09-03
Layout improvements #2

MantasAD:
Improved map upload page to be more responsive.
Added drag and drop sorting to map credits lists.
Added ability to select multiple files at once.
Made zones file upload to be mandatory.
Improved news and Twitch streams pages to be more responsive.
Cleared up layout component and changed some global styles.
Other small fixes and improvements.


client/src/app/pages/not-found/main/not-found-main.component.html
```diff
@@ -1,3 +1,4 @@
+<nb-layout style="display:none"></nb-layout>
```
Gocnak: I remember Nebular requiring `nb-layout` to have its components working, is that what this is doing?

MantasAD: ```nb-layout``` component hides spinner on initialization, so that's why I'm using it.

PR #403, opened 2020-05-05
Feature: Implement docker

### LINKED ISSUE #402: Implement docker for building and running the server
Needs the following features:

- [x] Add a Dockerfile, that builds the client, then builds the server, and finally moves `client/build` into `server/public/`. 
- [x] Add a GitHub action on pushes to staging/master to build the Dockerfile and push to a registry (GitHub Packages/Docker hub).
- [x] Mount the persistent data (`runs`, `maps` and `img`) to the volume automatically when the container runs. 

Should the host machine listen for pushes to the registry and automatically pull the image, and run it, or should this be a manual process?

TheRealHona:
Closes #402 

This PR is a draft, as paths for persistent mounting needs to be decided.

Before this is merged, the org needs to setup GitHub packages for pushing the built images with GitHub Actions to. In addition to this, the repo needs to add a secret `GITHUB_TOKEN` for authenticating the push to GitHub packages.

### DIFF
--- a/.github/workflows/dockerimage.yml
+++ b/.github/workflows/dockerimage.yml
@@ -7,9 +7,9 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@master
+    - uses: actions/checkout@v2
     - name: Publish to Registry
-      uses: momentum-mod/website@staging
+      uses: elgohr/Publish-Docker-Github-Action@2.14
       with:
         name: momentum-mod/website/mmod-website-production
         username: ${{ github.actor }}
--- a/.github/workflows/dockerimage.yml
+++ b/.github/workflows/dockerimage.yml
@@ -1,0 +1,14 @@
+name: Publish Docker
+on: [push]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - name: Publish to Registry
+      uses: TheRealHona/website@staging
+      with:
+        name: momentum-mod/website/mmod-website-production
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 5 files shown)


.github/workflows/dockerimage.yml
```diff
@@ -0,0 +1,14 @@
+name: Publish Docker
+on: [push]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - name: Publish to Registry
+      uses: TheRealHona/website@staging
```
Gocnak, line 9: Wouldn't this need to change to be our repo?

.github/workflows/dockerimage.yml
```diff
@@ -0,0 +1,14 @@
+name: Publish Docker
+on: [push]
```
Gocnak, line 2: Shouldn't this be on push to the staging branch?

env-vars.list
```diff
@@ -0,0 +1,32 @@
...
+# The Discord Client ID and Secret for using their API
+DISCORD_CLIENT_ID=
+DISCORD_CLIENT_SECRET=
+
+# The Twitch Client ID and Secret for using their API
+TWITCH_CLIENT_ID=
+TWITCH_CLIENT_SECRET=
+
+# A generated secret for express' sessions
+EXPRESS_SESSION_SECRET=
```
Gocnak, line 32: How does this file work, do you need to make a copy of it or should the prod server just fill this in before running `runDocker.sh`?

TheRealHona: The `docker run` docker command mounts it with `--env-file env-vars.list`. So yep, you just need to fill it in before running.

TheRealHona:
Changed it to only build on pushes to staging and use momentum-mod/website not TheRealHona/website.

Issue #237, opened 2018-12-09
Add linter to the server


Corey28:

TheRealHona:
We can add ESLint, and configure it to the current style, then add a github action to run the linter as a check, both on staging + PRs

PR #432, opened 2020-08-06
Implement Bunyan + Seq logging

### LINKED ISSUE #410: Replace logging with Bunyan + Seq
In an effort to provide a centralised logging solution, the discord bot has move to Seq logging, and the website should too.

https://docs.datalust.co/docs/using-nodejs

TheRealHona:
Closes #410 

Awaiting testing

### DIFF
--- a/server/config/express.js
+++ b/server/config/express.js
@@ -1,6 +1,5 @@
 const express = require('express'),
 	path = require('path'),
-	logger = require('morgan'),
 	compress = require('compression'),
 	methodOverride = require('method-override'),
 	passport = require('passport'),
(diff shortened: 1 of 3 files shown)


server/config/express.js
```diff
@@ -146,7 +167,7 @@ module.exports = (app, config) => {
 		app.use((err, req, res, next) => {
 			const status = err.status || (isCelebrate(err) ? 400 : 500);
 			if (status === 500 || isCelebrate(err))
-				console.error(err);
```
Gocnak, line 149: For the test environment, we should probably have this still be console. Tests are only run locally or in CI, not on production.

Gocnak:
I also noticed some discrepancies with spacing, the server uses tabs. When the server linter (#237 ) happens, this can change, but for now stay consistent with tabs.

server/config/express.js
```diff
@@ -17,17 +16,39 @@ const express = require('express'),
...
+			level: 'info'
+			})
+		]
+	})
+	
+	if (app.get('env') === 'development' || app.get('env') === 'production') {
+		app.use(bunyanMiddleware(
+			{
+				logger: logger, 
+			}));
```
Gocnak, line 51: This can be one line

server/config/express.js
```diff
@@ -17,17 +16,39 @@ const express = require('express'),
...
+		name: 'Website',
+		streams: [
+			{
+				stream: process.stdout,
+				level: 'warn'
+			},
+			seq.createStream({
+			serverUrl: process.env.SEQ_ADDRESS,
+			apiKey: process.env.SEQ_TOKEN,
+			level: 'info'
```
Gocnak, line 42: Indent once more for easier readability

server/config/express.js
```diff
@@ -17,17 +16,36 @@ const express = require('express'),
...
+		streams: [
+			{
+				stream: process.stdout,
+				level: 'warn',
+			},
+			seq.createStream({
+				serverUrl: process.env.SEQ_ADDRESS,
+				apiKey: process.env.SEQ_TOKEN,
+				level: 'info',
+			})
```
Gocnak, line 43: Also, what happens if people don't have these defined? Is Seq now a requirement for development or does this gracefully handle it?

server/config/express.js
```diff
@@ -17,17 +16,63 @@ const express = require('express'),
...
 const swaggerSpec = swaggerJSDoc({
 	swaggerDefinition: swaggerDefinition,
 	apis: ['./docs/**/*.yaml'],
 });
 
+function wrappedStdout() {
+	return {
+		write: (entry) => {
+			// Format console output to be human readable
+			var logObject = JSON.parse(entry);
```
Gocnak, line 33: Var is pretty bad to use, use `let` or `const`, in this case it's `const` as `logObject` never gets re-assigned to.
```suggestion
			const logObject = JSON.parse(entry);
```

TheRealHona:
Yep, I will rebase this and change to `const`

server/config/express.js
```diff
@@ -17,17 +16,63 @@ const express = require('express'),
...
 	swaggerDefinition: swaggerDefinition,
 	apis: ['./docs/**/*.yaml'],
 });
 
+function wrappedStdout() {
+	return {
+		write: (entry) => {
+			// Format console output to be human readable
+			const logObject = JSON.parse(entry);
+			severity = bunyan.nameFromLevel[logObject.level].toUpperCase();
```
Gocnak, line 34: Wait I just noticed this, doesn't this need a `const` as well?

PR #434, opened 2020-08-21
Fix #401 w/ javascript on (click) action

### LINKED ISSUE #401: View jumps to header when clicking about/gamemodes/links for the first time
**Description**
When the page is first loaded, clicking the about/gamemodes/links on the nav for the first time always returns the page to the header. Subsequent clicks sends the view to the right section correctly.

This happens regardless of whether cache is refreshed.

**Browser**
Seems to happen on every browser & resolution. Tested on firefox/chrome/edge.

davis-matthew:
Fixes #401 

This issue had two potential fixes:

> **1. Make the href double activate to ensure it jumps down to the element** 
> 
> 
> **Pros:**
> - Keeps the current functionality while fixing the bug exactly (href anchor will be appended to the url & the button clicks will function correctly)
> 
> **Cons:**
> - The browser back button will take two clicks to go back
> - The browser will _almost_ definitely refresh the page. This is slightly jarring for a button which just scrolls down the page
> - Could potentially cause the "you are leaving the website" popup to appear because it is technically redirecting when running the script to perform the double click

> **2. Change functionality to the on-click "(click)" action and a corresponding function call** 
> 
> **Pros:**
> - Works every time without reloading the page
> - The overridden href can be left in so pre-existing links (ex. http://momentum-mod.org/#gamemodes) will still work and scroll to the correct position. This also ensures the cursor will still change to the hand icon
> 
> **Cons:**
> - Will not append the href anchor tag to the url (stays as http://momentum-mod.org/)
> 

I did the second fix as it seemed like the better option. The href tag not being on the url is slightly annoying but could be remedied by a feature like a mouseover activated "copy link" button which copies the corresponding link for sharing a specific section (this is a common fix)

### DIFF
--- a/client/src/app/pages/main/nav/nav.component.ts
+++ b/client/src/app/pages/main/nav/nav.component.ts
@@ -56,16 +56,8 @@ export class NavComponent implements OnInit {
     });
   }
 
-  scrollTo(elementID:string): boolean {
-    switch(elementID){
-      case 'about':
-      case 'gamemodes':
-      case 'footer':
-        document.getElementById(`${elementID}`).scrollIntoView();
-        break;
-      default:
-        console.log("Unexpected Navigation Location");
-    }
+  scrollTo(elementID : string): boolean {
+    document.getElementById(`${elementID}`).scrollIntoView({ behavior: 'smooth' });
     return false;
   }
 }
--- a/client/src/app/pages/main/nav/nav.component.ts
+++ b/client/src/app/pages/main/nav/nav.component.ts
@@ -55,4 +55,17 @@ export class NavComponent implements OnInit {
       }
     });
   }
+
+  scrollTo(elementID:string): boolean {
+    switch(elementID){
+      case 'about':
+      case 'gamemodes':
+      case 'footer':
+        document.getElementById(`${elementID}`).scrollIntoView();
+        break;
+      default:
+        console.log("Unexpected Navigation Location");
+    }
+    return false;
+  }
 }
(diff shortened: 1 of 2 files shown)


client/src/app/pages/main/nav/nav.component.ts
```diff
@@ -55,4 +55,17 @@ export class NavComponent implements OnInit {
       }
     });
   }
+
+  scrollTo(elementID:string): boolean {
+    switch(elementID){
```
Gocnak, line 60: I don't think the switch is necessary here, it's only used by three elements.

client/src/app/pages/main/nav/nav.component.ts
```diff
@@ -55,4 +55,17 @@ export class NavComponent implements OnInit {
...
+      case 'about':
+      case 'gamemodes':
+      case 'footer':
+        document.getElementById(`${elementID}`).scrollIntoView();
```
Gocnak, line 64: Let's pass in
```suggestion
        document.getElementById(`${elementID}`).scrollIntoView({ behavior: 'smooth' });
```
It defaults to auto, but I'm not sure auto would be smooth here.

client/src/app/pages/main/nav/nav.component.ts
```diff
@@ -55,4 +55,17 @@ export class NavComponent implements OnInit {
       }
     });
   }
+
+  scrollTo(elementID:string): boolean {
```
Gocnak, line 59:
```suggestion
  scrollTo(elementID : string): boolean {
```

PR #440, opened 2020-09-19
Adding front end CI building

arh78750:
just very simple use of github actions building the frontend and uploading it as an artifact

Right now node_modules and the node install are cached between runs but no parts of the angular build are cached


arh78750:
also fixed a small linting error preventing me from building

TheRealHona:
Looks good, we might look into automatic uploads to an S3 Bucket for production.

arh78750:
yeah that should be really easy to setup in the future

PR #444, opened 2020-09-30
Update README.md

sayyamjain78:
more useful heading


Gocnak:
No not really. If you're going for a hacktober T shirt you're going to need to do something more meaningful instead of spamming repos with meaningless readme changes.

PR #445, opened 2020-09-30
don't show to much 

SahilDiwakar:


Gocnak:
Ah yes, why would anyone want to know the technology used in our website... 🙄

PR #449, opened 2020-10-01
Updated env-vars.list

SumitRodrigues:


TheRealHona (commented):
Nothing uses the `NAME` env var, what is the point of this PR?

PR #459, opened 2020-12-31
Added infinite list for recent activities

vindow:
For issue #200 

I used Nebular's built-in infinite list functionality. 

I wasn't sure what threshold I should have picked (0 is when scroll bar all the way down). The emitter for the activity-list will also re-emit if you keep scrolling to the bottom and back up when there are no more activities left to load but the function to handle the emit in activity-card will immediately return and not call the API if the list is exhausted. I also had to add a check that it would only load more for recent activity since I noticed activity-card was reused for followed activities. Lastly, I had to make a slight adjustment to the activity service so that I could call older activities from the API.

### DIFF
--- a/client/src/app/@core/data/activity.service.spec.ts
+++ b/client/src/app/@core/data/activity.service.spec.ts
@@ -50,7 +50,7 @@ describe('ActivityService', () => {
     });
     it('#getRecentActivity() should return recent Activity', () => {
       httpClientSpy.get.and.returnValue(of(expectedActivities));
-      activityService.getRecentActivity().subscribe(value =>
+      activityService.getRecentActivity(0).subscribe(value =>
           expect(value).toEqual(expectedActivities, 'expected activity'),
         fail,
       );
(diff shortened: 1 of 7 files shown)


vindow:
So I noticed the check failed but it was because I changed an activity service function to be able to request older activities. Is there something I should do on my end? I'm not really experienced in working on non-solo projects.

Hona:
@vindow `client/src/app/@core/data/activity.service.spec.ts` on line 53, change `getRecentActivity()` to `getRecentActivity(0)`

Gocnak:
Yep just looks like you missed a spot with defining the offset. What Hona suggested is what you need to do to fix it!

vindow:
Okay, I tried supplying NbLayoutScrollService as a provider in the Theme module locally to fix the current error in test (not sure why I wasn't running tests locally before this...) but it still complains that there is no provider supplied when I run npm run test. I feel like I'm missing something obvious here.

MantasAD:
Make sure to import `NbLayoutScrollService` into [`ActivityListComponent` testing module](https://github.com/momentum-mod/website/blob/4a4b0d1b2e2f14f02e13fbd87e5dd46fcd3dd38d/client/src/app/%40theme/components/activity/activity-list/activity-list.component.spec.ts#L15).

vindow:
@MantasAD Adding that + another import it needed worked. Thanks!

Gocnak (approved):
Looks good to me! You can PM me on discord for a key if you don't already have one. Thanks for the contribution and looking forward to more! :)

Issue #252, opened 2018-12-10
Hook up Run History filters


Gocnak:
Currently there just not hooked up

vindow:
So if you apply the filters client-side you'll run into some issues with pagination and filtering. You can get a page of no results since the search result might be on the next page of 10 unfiltered results. Here's a picture example: 

![example](https://user-images.githubusercontent.com/5974776/103715459-8a00ac80-4f75-11eb-8a97-43d81aa5c23d.png)

Would it be better to just re-query from the backend with filter params (probably requires some backend reworking) so you can get a list of 10 filtered results per page? This issue somewhat applies to other lists of things too (such as activities) but it's more noticeable with the pagination of the user runs list.

Gocnak:
Yeah it would be better to just re-query upon applying the filters. Could have them auto-apply with a debounce like the searching does, or just for now, use the Apply button and see if people think that's good enough.

PR #463, opened 2021-01-10
Enabled run history filters for user pages

vindow:
For issue #252 

### Backend Changes
I'm not entirely sure on the implementation for the backend on this one. I had to change the run model to be able to take in the isPB, mapName, and order parameters for the filtering to work. I left a few comments with questions on whether the structure was correct/good practice for isPB and order. I'm also unsure whether mapName is implemented correctly since I'm not too familiar with working in sequelize/SQL (it's been a long time since I've touched SQL in general).

Additionally I needed to add these parameters into validation and I tried reusing what I could that was already in query-param.js. I had to let search accept an empty string which I'm not sure if that could have unintended side-effects elsewhere.

### Frontend Changes
I elected to just use an apply button for now for the filter. I tried to keep it on the same line for most resolutions but the button does make a new row at certain breakpoints before everything goes into their own row. There was/is a minor issue with the responsive styling at 1200-1237px width where the is PB label uses a new line seen here:
image
Not sure how to fix that though.

As for functionality I simply had the loadRunHistory function pass in the parameters based on the form group values when querying. I also had to make the button call a new onFilterApply function so that if the user changes the filter and applies the page knows to start from the first page of results again.

### DIFF
--- a/client/src/app/pages/dashboard/profile/profile-run-history/profile-run-history.component.ts
+++ b/client/src/app/pages/dashboard/profile/profile-run-history/profile-run-history.component.ts
@@ -61,9 +61,9 @@ export class ProfileRunHistoryComponent implements OnInit {
     this.usersService.getRunHistory(this.user.id, {
       params: {
         expand: 'map',
-        mapName: this.filterFG.value.map,
-        isPB: this.filterFG.value.isPersonalBest,
-        order: this.filterFG.value.order,
+        mapName: this.currentFilter.map,
+        isPB: this.currentFilter.isPersonalBest,
+        order: this.currentFilter.order,
         limit: this.pageLimit,
         offset: (this.currentPage - 1) * this.pageLimit,
       },
@@ -83,8 +83,8 @@ export class ProfileRunHistoryComponent implements OnInit {
     // Some destructuring to shorten the upcoming if statement
     const { isPersonalBest, map, order } = this.filterFG.value;
     if ( // Don't do anything if the filters didn't change
-      this.currentFilter.isPersonalBest === isPersonalBest ||
-      this.currentFilter.map === map ||
+      this.currentFilter.isPersonalBest === isPersonalBest &&
+      this.currentFilter.map === map &&
       this.currentFilter.order === order
     )
       return;
--- a/client/src/app/pages/dashboard/profile/profile-run-history/profile-run-history.component.ts
+++ b/client/src/app/pages/dashboard/profile/profile-run-history/profile-run-history.component.ts
@@ -22,6 +22,11 @@ export class ProfileRunHistoryComponent implements OnInit {
   currentPage: number;
   runCount: number;
   showFilters: boolean;
+  currentFilter: {
+    isPersonalBest: boolean,
+    map: string,
+    order: string,
+  };
 
   filterFG: FormGroup = this.fb.group({
     'isPersonalBest': [false],
@@ -38,6 +43,11 @@ export class ProfileRunHistoryComponent implements OnInit {
     this.runCount = 0;
     this.showFilters = false;
     this.runHistory = [];
+    this.currentFilter = {
+      isPersonalBest: this.filterFG.value.isPersonalBest,
+      map: this.filterFG.value.map,
+      order: this.filterFG.value.order,
+    };
   }
 
   ngOnInit() {
@@ -46,10 +56,14 @@ export class ProfileRunHistoryComponent implements OnInit {
       this.loadRunHistory();
     });
   }
+
   loadRunHistory() {
     this.usersService.getRunHistory(this.user.id, {
       params: {
         expand: 'map',
+        mapName: this.filterFG.value.map,
+        isPB: this.filterFG.value.isPersonalBest,
+        order: this.filterFG.value.order,
         limit: this.pageLimit,
         offset: (this.currentPage - 1) * this.pageLimit,
       },
@@ -62,6 +76,20 @@ export class ProfileRunHistoryComponent implements OnInit {
 
   onPageChange(pageNum: number) {
     this.currentPage = pageNum;
+    this.loadRunHistory();
+  }
+
+  onFilterApply() {
+    // Some destructuring to shorten the upcoming if statement
+    const { isPersonalBest, map, order } = this.filterFG.value;
+    if ( // Don't do anything if the filters didn't change
+      this.currentFilter.isPersonalBest === isPersonalBest ||
+      this.currentFilter.map === map ||
+      this.currentFilter.order === order
+    )
+      return;
+    this.currentFilter = { isPersonalBest, map, order };
+    this.currentPage = 1; // Reset page back to 1 so you don't potentially pull a page past the last filtered page
     this.loadRunHistory();
   }
 }
--- a/server/src/controllers/users.js
+++ b/server/src/controllers/users.js
@@ -72,6 +72,9 @@ module.exports = {
 
 	getRuns: (req, res, next) => {
 		req.query.playerID = req.params.userID;
+		if (!req.query.order) {
+			req.query.order = 'date';
+		}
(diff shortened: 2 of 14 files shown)


client/src/app/pages/dashboard/profile/profile-run-history/profile-run-history.component.ts
```diff
@@ -64,4 +78,18 @@ export class ProfileRunHistoryComponent implements OnInit {
...
+
+  onFilterApply() {
+    // Some destructuring to shorten the upcoming if statement
+    const { isPersonalBest, map, order } = this.filterFG.value;
+    if ( // Don't do anything if the filters didn't change
+      this.currentFilter.isPersonalBest === isPersonalBest ||
+      this.currentFilter.map === map ||
+      this.currentFilter.order === order
+    )
+      return;
```
Gocnak, line 90: This logic doesn't seem correct. If you change one but not the other(s), wouldn't this return true and not search with the new query?

client/src/app/pages/dashboard/profile/profile-run-history/profile-run-history.component.ts
```diff
@@ -46,10 +56,14 @@ export class ProfileRunHistoryComponent implements OnInit {
...
     });
   }
+
   loadRunHistory() {
     this.usersService.getRunHistory(this.user.id, {
       params: {
         expand: 'map',
+        mapName: this.filterFG.value.map,
+        isPB: this.filterFG.value.isPersonalBest,
+        order: this.filterFG.value.order,
```
Gocnak, line 66: Shouldn't this be using `this.currentFilter` ?

server/src/models/run.js
```diff
@@ -727,6 +729,18 @@ module.exports = {
...
+		}
+		if (queryParams.mapName) {
+			queryOptions.include.push({
+				model: Map,
+				as: 'map',
+				where: { name: {[Op.startsWith]: queryParams.mapName} }
+			});
+		}
+		if (queryParams.isPB) {
+			queryOptions.include[1].required = queryParams.isPB; // Works, but feels wrong to assume the array index directly like this
```
Gocnak, line 743: I believe you could instead put this directly in the include object, with it like:
```
					attributes: ['rank'],
					required: queryParams.isPB ?? false,
```

server/src/models/run.js
```diff
@@ -727,6 +729,18 @@ module.exports = {
...
 			if (queryParams.order === 'date')
 				queryOptions.order = [['createdAt', 'DESC']];
+			if (queryParams.order === 'time')
+				queryOptions.order = [[sequelize.literal('ticks * tickRate'), 'ASC']];
+		}
+		if (queryParams.mapName) {
+			queryOptions.include.push({
+				model: Map,
+				as: 'map',
+				where: { name: {[Op.startsWith]: queryParams.mapName} }
```
Gocnak, line 739: We will want to use name matching instead of `startsWith`. See [how the map model does it here](https://github.com/momentum-mod/website/blob/646f8ab8f3eeb6d85d333c2b8266165166b6c4df/server/src/models/map.js#L171).

server/src/models/validation/query-param.js
```diff
@@ -4,9 +4,10 @@ const Joi = require('@hapi/joi');
 module.exports = {
 	limit: Joi.number().integer().min(1).max(20),
 	offset: Joi.number().integer().min(0).max(5000),
-	search: Joi.string(), // hmmm
+	search: Joi.string().allow(''), // Is there anywhere that this is used that shouldn't allow empty strings?
 	expand: Joi.string().regex(/[a-zA-Z,]*/),
 	priority: Joi.boolean(),
 	status: Joi.string().regex(/[0-9,]*/), // should probably be moved out to somewhere else? change to statusIn?
 	playerIDs: Joi.string().regex(/[0-9,]*/),
+	order: Joi.string().valid("date", "time").required(), // Should this be moved somewhere else? Should it accept more than just date and time?
```
Gocnak, line 12: Use single apostrophe for strings:
```suggestion
	order: Joi.string().valid('date', 'time').required(), // Should this be moved somewhere else? Should it accept more than just date and time?
```

server/src/models/run.js
```diff
@@ -708,9 +710,9 @@ module.exports = {
 					as: 'rank',
 					attributes: ['rank'],
 					required: false,
-				}
+				},
 			],
-			order: [['ticks', 'ASC']],
+			order: [['ticks', 'ASC']], // Should the order default to date instead of ticks?
```
Gocnak, line 715: Yeah it can default to date.

server/src/controllers/users.js
```diff
@@ -72,7 +72,6 @@ module.exports = {
 
 	getRuns: (req, res, next) => {
 		req.query.playerID = req.params.userID;
-		req.query.order = 'date';
```
Gocnak, line 75: I'm not sure the consequences of doing this but honestly if the default runMdl.getAll just sorts by date anyways this would be fine. Would be worth looking into what all uses `runMdl.getAll` and see if they rely on the order being `ticks`.

Gocnak (changes requested):

client/src/app/pages/dashboard/profile/profile-run-history/profile-run-history.component.ts
Gocnak: Which in this case you'd just want to change the `||` to `&&` to verify *nothing* changed, as opposed to *something* which it currently checks for.

server/src/controllers/users.js
vindow: It seems to only be used [here](https://github.com/momentum-mod/website/blob/646f8ab8f3eeb6d85d333c2b8266165166b6c4df/server/src/controllers/runs.js) elsewhere. I could also re-add the default query order parameter for date and override it only if an order parameter is issued to the service if that's more clear/safe.

client/src/app/pages/dashboard/profile/profile-run-history/profile-run-history.component.ts
vindow: Forgot to flip `||` and `&&' when I flipped the boolean logic. Fixed in 97e96fe

client/src/app/pages/dashboard/profile/profile-run-history/profile-run-history.component.ts
vindow: It might matter in edge cases that it isn't `this.currentFilter` (as in it might not work right as `this.filterFG.value` but even if it doesn't it makes more sense and is cleaner to make it so. Changed in 97e96fe

server/src/models/run.js
vindow: JS didn't like `??` and would throw an `Uncaught SyntaxError: Unexpected token '?'` but using `||` should work in this case since if `isPB` is false and not null it would still return the expected false result. Changed in b1bd918

server/src/models/run.js
vindow: Defaulted in b1bd918

server/src/models/validation/query-param.js
vindow: Fixed in b1bd918

server/src/models/run.js
vindow: Fixed in 2f43ccd. Let me know if it still doesn't look right.

server/src/controllers/users.js
vindow: On further review only the runs service seems to call `/api/runs/` from the aforementioned link and nothing seems to call the runs service `getRuns`. Nevertheless I've changed it back to be like the old result on default return in case something in the future relies on `getRuns` returning in time order. I did, however, make it return in time order instead of tick order since I presume that was what was expected from before.

Gocnak (approved):

Issue #258, opened 2018-12-10
Parse external Steam API request errors and return them properly


Gocnak:

TheRealHona:
Were there specific parts in the current server code this is referencing?

tsa96:
Nest backend does this: https://github.com/momentum-mod/website/blob/main/apps/backend/src/app/modules/steam/steam.service.ts

PR #470, opened 2021-02-01
Added error messages for external Steam API request errors

Devwulf:
Partly solves issue #258 

For the `models/user.js`, I handled the case and added an error message for when, given a list of steam ids, no players were found. I also added error messages for when the response code is not 200. 

As for the `controllers/auth.js`, I could neither find how the response object looks like for cases when `params.result !== 'OK'`, nor do I know how to generate a user ticket to test this API endpoint on Postman with. The docs for that API endpoint doesn't show the structure of the response either, so I really won't be able to properly parse `sres.data`. However, I found that if I passed in an invalid ticket value (like 0), the response body would look something like this instead:
```
{
    "response": {
        "error": {
            "errorcode": 3,
            "errordesc": "Invalid parameter"
        }
    }
}
```
For those cases, I just sent a Bad Request instead.

### DIFF
--- a/server/src/controllers/auth.js
+++ b/server/src/controllers/auth.js
@@ -88,12 +88,11 @@ module.exports = {
 					ticket: userTicket
 				}
 			}).then((sres) => {
-				const params = sres.data.response.params;
-				if (!params)
-					res.sendStatus(400); // Bad request, it'll have an error object instead
+				if (sres.data.response.error)
+					res.sendStatus(400).json(sres.data.response.error); // Bad request, and pass the error object as reason
 
-				if (params.result === 'OK') {
-					if (idToVerify === params.steamid) {
+				if (sres.data.response.params.result === 'OK') {
+					if (idToVerify === sres.data.response.params.steamid) {
 						user.findOrCreateFromGame(idToVerify).then((usr) => {
 							auth.genAccessToken(usr, true).then(token => {
 								res.json({
--- a/server/src/controllers/auth.js
+++ b/server/src/controllers/auth.js
@@ -89,7 +89,7 @@ module.exports = {
 				}
 			}).then((sres) => {
 				if (sres.data.response.error)
-					res.sendStatus(400).json(sres.data.response.error); // Bad request, and pass the error object as reason
+					res.sendStatus(400); // Bad request, and not sending the error object just in case of hidden bans
 
 				if (sres.data.response.params.result === 'OK') {
 					if (idToVerify === sres.data.response.params.steamid) {
--- a/server/src/controllers/auth.js
+++ b/server/src/controllers/auth.js
@@ -88,8 +88,12 @@ module.exports = {
 					ticket: userTicket
 				}
 			}).then((sres) => {
-				if (sres.data.response.params.result === 'OK') {
-					if (idToVerify === sres.data.response.params.steamid) {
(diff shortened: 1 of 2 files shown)


server/src/controllers/auth.js
```diff
@@ -88,8 +88,12 @@ module.exports = {
 					ticket: userTicket
 				}
 			}).then((sres) => {
-				if (sres.data.response.params.result === 'OK') {
-					if (idToVerify === sres.data.response.params.steamid) {
+				const params = sres.data.response.params;
+				if (!params)
```
Hona, line 92: This works, however I'd say to make it:

```node
if (sres.data.error)
```

Just to be a bit more readable and safer.

server/src/controllers/auth.js
```diff
@@ -88,8 +88,12 @@ module.exports = {
...
+					res.sendStatus(400); // Bad request, it'll have an error object instead
```
Hona, line 93: I think it is safe to respond with the error reasons by simply passing the error:

Maybe if there is a silent ban or something maybe not, needs @Gocnak 's opinion

```node
res.sendStatus(400).json(sres.data.response.error);
```

It could be prettied into a renamed format, but I think it is fine

Hona (changes requested):
Looks good so far! Just a few changes IMO

Hona (approved):

server/src/controllers/auth.js
Gocnak: As per the docs, [`sendStatus(status)`](https://expressjs.com/en/4x/api.html#res.sendStatus) does internal response handling, you would instead need to do
```
res.status(400).json(sres.data.response.error);
```

As per the guide on [`response.json()`](https://expressjs.com/en/4x/api.html#res.json)

Hona: Are we happy to respond to clients with the steam error message though? Would it let them know of hidden bans or something like that?

Gocnak: Yes good point. Let's just make this an error 400 here for now, as the code originally had.

server/src/controllers/auth.js
```diff
@@ -88,6 +88,9 @@ module.exports = {
 					ticket: userTicket
 				}
 			}).then((sres) => {
+				if (sres.data.response.error)
+					res.sendStatus(400).json(sres.data.response.error); // Bad request, and pass the error object as reason
```
Gocnak, line 92: Hey just noting ([as from before](https://github.com/momentum-mod/website/pull/470#discussion_r585250245)) this is still needing to change to be `status(400).json(...)`.
```suggestion
					res.status(400).json(sres.data.response.error); // Bad request, and pass the error object as reason
```

Gocnak (changes requested):

server/src/controllers/auth.js
Devwulf: Are we still sending the error object, or should we revert this to before I changed things?

Gocnak: Oh right. Then the correct thing to do here would be:
```suggestion
					res.sendStatus(400); // Bad request
```

Gocnak (approved):
Looks great! PM me on Discord for a key, sorry about the delay in reviewing!

Issue #415, opened 2020-07-06
Update map credits types


Gocnak:
Upon further reconsideration, the map credits types need some updating.

The AUTHOR type stays and becomes "anyone who worked on the map". It is implied with the order of the credits that the "top-most" name means the most important. For example, collaboration maps would list anyone who made a stage under this credit type.

CO_AUTHOR gets changes to become CONTRIBUTOR. These are people that gave assets, did very small work on the map, but still did work regardless to make the map. For example, people who made models out of the ramps, or provided original assets like textures to use for the map could be given this credit type.

TESTER stays exactly the same, and will be used to credit any testing whatsoever.

SPECIAL_THANKS stays exactly the same as well, and this is used for any miscellaneous credits, things like motivation or monetary support, etc.

On top of these changes, the credit entry should be updated to take in a small text field (varchar 255 should be good) for further explanation. This allows authors to be attributed to exactly what they made, or contributors for what exactly they contributed. For example, on a collaboration map, "author1" made stages 1, 3, and 5, so this text could be "Stages 1, 3, 5".

tsa96:
We should do some general cleanup of credits during this.
- There's some duplicate code in the various map credits related Angular components
- Currently during map credits updates the frontend sends PATCHs for *every* credit, even if unchanged.

tsa96:
This we absolutely want before map submission, so getting started on my current branch. Note we no longer use `varchar(x)` (not perf benefit over `text` in Postgres.

PR #471, opened 2021-02-04
Added ability to filter uploaded, favorited, in-library, and general maps

### LINKED ISSUE #265: Be able to filter uploaded maps
Add filter controls inside of the map upload status to easily filter results by name, status, etc

vindow:
Closes #265

Notable things:
- I added some commented out code in the frontend that will eventually need to be wired up to the backend when #415 is closed.
- The design for the search bar should be responsive at all the breakpoints (and between them). There are some slight tweaks that need to be made when the author search is added and I've added comments on what needs to be changed.
- The dropdown lists are populated based on the `MapUploadStatus` and `MomentumMapType` enums, but the proper display strings need to be added in `formatStatusEnum()` and `formatTypeEnum()` when new statuses/gamemodes are added. Same thing with `map-list-item-component` when a new status is added.
- I added different no map found messages for when there are no results but a search field was selected to reflect that there might be results with different search parameters.
- I updated the status descriptions on the `map-list-item` to be formatted without underscores.
- I changed onPageChange to just update `currentPage` and then recall `loadMap()`. I'm not entirely sure why the old function was written to call the router.
- The backened needed to be adjusted a little bit to accept `type` as a query parameter.


Hona:
Don't worry about the server tests failing, I reran it and it passes now. Theres a bit of a bug with the tests

client/src/app/pages/dashboard/maps/map-list/map-list.component.ts
```diff
@@ -136,4 +144,27 @@ export class MapListComponent implements OnInit {
...
+        return 'Public Testing';
+      case MapUploadStatus.READY_FOR_RELEASE:
+        return 'Ready for Release';
+      case MapUploadStatus.REJECTED:
+        return 'Rejected';
+      case MapUploadStatus.REMOVED:
+        return 'Removed';
+      default:
+        return this.statusEnum[key];
+    }
```
Gocnak: Could this exist in the MapUploadStatus interface as a method?

client/src/app/pages/dashboard/maps/map-list/map-list-item/map-list-item.component.ts
```diff
@@ -21,18 +21,48 @@ export class MapListItemComponent implements OnInit {
...
+        break;
+      case MapUploadStatus.REJECTED:
+        this.status = 'Rejected';
+        break;
+      case MapUploadStatus.REMOVED:
+        this.status = 'Removed';
+        break;
+      default:
+        this.status = this.statusEnum[this.map.statusFlag];
+    }
```
Gocnak, line 65: This should use the same code that map-list-component ends up using.

client/src/app/pages/dashboard/maps/map-list/map-list.component.ts
```diff
@@ -164,7 +195,36 @@ export class MapListComponent implements OnInit {
...
+      case MomentumMapType.BHOP:
+        return 'Bunny Hop';
+      case MomentumMapType.KZ:
+        return 'Climb (KZ/XC)';
+      case MomentumMapType.RJ:
+        return 'Rocket Jump';
+      case MomentumMapType.SJ:
+        return 'Sticky Jump';
+      case MomentumMapType.TRICKSURF:
+        // Is it Tricksurf, TrickSurf, or Trick Surf?
```
Gocnak, line 219: Just Tricksurf for now.

client/src/app/pages/dashboard/maps/map-list/map-list.component.ts
```diff
@@ -136,4 +179,61 @@ export class MapListComponent implements OnInit {
...
+      case MapUploadStatus.REJECTED:
+        return 'Rejected';
+      case MapUploadStatus.REMOVED:
+        return 'Removed';
+      default:
+        return MapUploadStatus[key];
+    }
+  }
+
+  formatTypeEnum(key: number) {
```
Gocnak, line 206: Could this exist in MomentumMapType interface as a helper method?

Gocnak (changes requested):

vindow:
I found a bug while testing the refactoring where changing the search parameters but not hitting search could change the no maps found message so I went ahead and fixed that as well in 2ca25a7. I had to store the last search and then refer to it in isSearchFiltered() instead of the current form object. I also changed `status` in `MapAPIQueryParams` to be number since it's an enum (`MapAPIQueryParams` is only referenced in map-list-component).

Gocnak:
Looks like this branch needs rebasing to fix the conflict in map-list.component.

client/src/app/pages/dashboard/maps/map-list/map-list.component.html
```diff
@@ -32,6 +56,14 @@ <h3 class="m-0" *ngSwitchCase="mapListType.TYPE_UPLOADS">
       <i>You have not uploaded any maps. Try <a routerLink="/dashboard/maps/uploads/new">uploading a new one!</a></i></h3>
   </div>
 </div>
+<div class="text-center" *ngIf="maps.length === 0 && requestSent && isSearchFiltered()">
+  <div [ngSwitch]="type">
+    <h3 class="m-0" *ngSwitchDefault>No maps with those search parameters found.</h3>
+    <h3 class="m-0" *ngSwitchCase="mapListType.TYPE_LIBRARY">No maps with those search parameters found in your library.</h3>
+    <h3 class="m-0" *ngSwitchCase="mapListType.TYPE_FAVORITES">No favorite maps with those search parameters found.</h3>
+    <h3 class="m-0" *ngSwitchCase="mapListType.TYPE_UPLOADS">You have not uploaded any maps with those search parameters.</h3>
```
Gocnak, line 64: This should happen in the component backing (Typescript) code, since the type of list is only set once.

Gocnak (changes requested):

client/src/app/pages/dashboard/maps/map-list/map-list-item/map-list-item.component.html
```diff
@@ -30,7 +30,7 @@
   <div class="col-12 col-lg-6">
     <div class="row no-gutters align-items-sm-center">
       <div class="col-12 col-sm mb-3 mb-sm-0 text-center" *ngIf="isUpload">
-        <span class="mapStatus">{{statusEnum[map.statusFlag]}}</span>
```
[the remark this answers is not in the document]
vindow: It seems to be used [here](https://github.com/momentum-mod/website/blob/897aa1264323f446dea7af966b26b8231ef0ef40/client/src/app/pages/dashboard/maps/map-list/map-list-item/map-list-item.component.html#L2) earlier in the file and the the scss file uses the enum names to draw the colored borders. Should I change the names [here](https://github.com/momentum-mod/website/blob/fa5fbdc03ceb2746a2eb71e62662dc6d7ad62844/client/src/app/pages/dashboard/maps/map-list/map-list-item/map-list-item.component.scss#L50) to be the properly cased names so I can drop statusEnum altogether?

client/src/app/pages/dashboard/maps/map-list/map-list.component.html
vindow: I refactored it into the component but there's a similar switch statement in the template file right before it here: https://github.com/momentum-mod/website/blob/897aa1264323f446dea7af966b26b8231ef0ef40/client/src/app/pages/dashboard/maps/map-list/map-list.component.html#L51-L57

Should I refactor that out as well?

vindow: Okay never mind I just realized it would require bypassing the sanitizer and would ultimately be messier.

Gocnak: Yeah it gets more complicated when there's another element in the middle. Typically if it's just text, though, you can definitely using a component variable binding.

client/src/app/pages/dashboard/maps/map-list/map-list-item/map-list-item.component.html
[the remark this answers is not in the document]
Gocnak: Yup!

client/src/app/pages/dashboard/maps/map-list/map-list.component.html
```diff
@@ -32,6 +56,9 @@ <h3 class="m-0" *ngSwitchCase="mapListType.TYPE_UPLOADS">
       <i>You have not uploaded any maps. Try <a routerLink="/dashboard/maps/uploads/new">uploading a new one!</a></i></h3>
   </div>
 </div>
+<div class="text-center" *ngIf="maps.length === 0 && requestSent && isSearchFiltered()">
+  <h3 class="m-0">{{getNoMapsFoundText()}}</h3>
```
Gocnak, line 60: What I meant here was to assign it to a local variable. Since bindings like this will be evaluated multiple times, this method would still be called when it should ideally know what type it is on `ngInit`, where a single string could be assigned and then here could reference it in interpolation.

Gocnak (changes requested):

Gocnak (approved):

PR #472, opened 2021-02-05
Changed UserID type from number to string on frontend for GUID support

vindow:
Seems okay for now. I'll have to double check it when the rewrite backend gets expanded more.


Hona:
Thanks for the contribution!

PR #475, opened 2021-03-01
Fixed pagination issue upon removal of last map in library and favorites

### LINKED ISSUE #473: Removing maps from a library page does not move to another page
If the library contains multiple pages, and you remove all maps from one page,  the list does not update with the previous page(s) and instead shows as if there are no more maps to display, with no UI elements to allow the user to switch pages
###### Removed all maps from page 2, the list does not update  with page 1:
image

Reloading does not move the user to another page

vindow:
Fixes #473

I added a check upon removal from library/favorites in `libraryUpdate()` and `libraryUpdate()` that checks for whether the user in on the last page + the last map of that page is being removed + the user isn't on page 1. Then all I needed to do was decrement the page count so it wouldn't query a page now out of index.

### DIFF
--- a/client/src/app/pages/dashboard/maps/map-list/map-list.component.ts
+++ b/client/src/app/pages/dashboard/maps/map-list/map-list.component.ts
@@ -121,13 +121,19 @@ export class MapListComponent implements OnInit {
   }
 
   libraryUpdate(): void {
-    if (this.type === MapListType.TYPE_LIBRARY)
+    if (this.type === MapListType.TYPE_LIBRARY) {
+      if (this.maps.length === 1 && this.currentPage * this.pageLimit >= this.mapCount && this.currentPage > 1)
+        this.currentPage -= 1;
       this.loadMaps();
+    }
   }
 
   favoriteUpdate() {
-    if (this.type === MapListType.TYPE_FAVORITES)
+    if (this.type === MapListType.TYPE_FAVORITES) {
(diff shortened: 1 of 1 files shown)


client/src/app/pages/dashboard/maps/map-list/map-list.component.ts
```diff
@@ -121,13 +121,19 @@ export class MapListComponent implements OnInit {
...
+      if (this.maps.length === 1 && this.currentPage * this.pageLimit >= this.mapCount && this.currentPage > 1)
+        this.currentPage -= 1;
       this.loadMaps();
+    }
   }
 
   favoriteUpdate() {
-    if (this.type === MapListType.TYPE_FAVORITES)
+    if (this.type === MapListType.TYPE_FAVORITES) {
+      if (this.maps.length === 1 && this.currentPage * this.pageLimit >= this.mapCount && this.currentPage > 1)
```
Hona, line 133: I'd like to see this duplicated check: `this.maps.length === 1 && this.currentPage * this.pageLimit >= this.mapCount && this.currentPage > 1` be moved to a function with a descriptive name

client/src/app/pages/dashboard/maps/map-list/map-list.component.ts
```diff
@@ -121,13 +121,19 @@ export class MapListComponent implements OnInit {
...
+        this.currentPage -= 1;
```
Hona, line 134: You may as well use `this.currentPage--`

Hona (changes requested):

Hona (approved):

PR #480, opened 2021-03-12
Added links to forums on sidebar and homepage footer

### LINKED ISSUE #479: Add link to forums
https://forum.momentum-mod.org/

Can use the `<nb-icon icon="people-icon">` icon from the [Eva Icons pack](https://akveo.github.io/eva-icons/#/?searchKey=people&type=fill).

vindow:
Closes #479

Let me know if the order of things should be changed in either area.


Gocnak (approved):
Looks good to me, if we ever get anal about how to place the icon we can revisit it later.

Issue #424, opened 2020-07-19
More CI improvements - caching builds


Hona:
From discussion with Gocnak, we decided to do the following to save some build time, as the build is ~15 minutes, and it would be nice to try to reduce this.

On PR:
- Travis:
   - Tests
- GitHub Action:
   - check cache for server/client stuff
   - install client dependencies
   - Build client:prod
   - yarn install server stuff?
   - cache server and client dependencies
   - output dist/ as artifact

On merge:
- GitHub Action (build Dockerfile):
   - Pull in cache for server dependencies
   - pull in client:prod build output (/dist/)
   - create local folders as Dockerfile sees fit for mounting
   - create Dockerfile with the server part of the Dockerfile as normal (build server packages using cache and mounted /dist/)

> Gocnak: And with a proper cache I'm expecting this to take no longer than 10 minutes combined, as the Travis build for the client only took 6 minutes, including caching things

Gocnak:
Status update on this:

#484 did a lot of great work to cache things, leading to the server tests taking under a minute, and client tests taking ~3-4 minutes. AngularBuild still happens (and catches things from time to time) but its output isn't used, which takes the longest time, ~8 mins.

Client "prod" build is now manually triggered and while it takes around 6 minutes, can't really do too much other than reduce dependencies in Angular somehow (same for AngularBuild).

Server docker file build takes under 2 minutes, which could be better with package caching there I think. Most of the time is actually building and publishing the dockerfile itself, so we would be shaving more of the `RUN yarn install` step there.


As we still have more work to do, this is pretty satisfactory for the time being. Maybe in the future (post backend TypeScript port) we can look into further caching gains, and make specific cards for the actions to take instead of monolithic mega ones like this one.

PR #484, opened 2021-04-03
Migrate tests to GitHub Actions

### LINKED ISSUE #465: Move Off of Travis CI
Their [new pricing scheme](https://travis-ci.com/plans) is abhorrent. Unfortunate change of leadership/ownership of their platform led to this. Move to GitHub actions, consider alternatives if Actions do not suffice for this repo.

strafe:
Closes #465.

I chose to emit the `Production Build` job as it is already covered by the new [`AngularBuild`](https://github.com/momentum-mod/website/blob/staging/.github/workflows/angularbuild.yml) action. A sample run of these workflows [can be seen on my fork](https://github.com/strafe/website/actions/runs/714690169).

[GitHub's Ubuntu runner environments have the default MySQL root user password set to `root`](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#mysql). If this ends up being a problem for running tests locally, a condition in the shell script is probably the best course.


strafe:
Also worth noting that I didn't include caching in this workflow, GitHub actions has [a pretty powerful caching action](https://github.com/actions/cache). But it's likely out of the scope of this particular PR and relates more to #424.

.github/workflows/test.yml
```diff
@@ -0,0 +1,63 @@
...
+env:
+  NODE_ENV: test
+
+jobs:
+  client-tests:
+    name: Client Tests
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [12.x]
```
Gocnak, line 20: Usually the matrix makes sense if we were some web library trying to be backwards compatible, but since our site will typically only use one version of node, it's safe to remove the matrix in favor of just one node version, in this case, 12.x.

Gocnak (changes requested):

.github/workflows/test.yml
strafe: @Gocnak I set it like this because that's how it appears in the existing `AngularBuild` workflow in this repo, I supposed it was because the site would be moving towards multiple versions soon so left it in. Should it be removed from one or both or none?

Gocnak:
Honestly you could add the caching to this PR if you'd like! Anything to optimize the build times haha.

.github/workflows/test.yml
Gocnak: They can be removed from both! I think it was left in the AngularBuild one by accident, we'll only have one node target yep.

.github/workflows/test.yml
```diff
@@ -0,0 +1,86 @@
...
+        cd ./client/
+        npm install -g @angular/cli && yarn install --prefer-offline
+        npm run test
+
+  server-tests:
+    name: Server Tests
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [12.x]
```
Gocnak, line 54: Missed one!

strafe: > Missed one!

Haha, was saving it to trigger a second build to check the cache had saved for client tests but went down a rabbit hole of commits anyway 👀

strafe:
Assuming this set of tests pass, this should be good to go. I also ended up setting the workflows to not trigger when editing just Markdown files because that's a waste of build minutes. This can be expanded upon in the future in theory if someone wanted to gather a list of files that don't affect the build. Docs: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths

Hona:
Very awesome work! Are you in our discord? 

We'd love to have you keep making PRs if that's something you'd be interested in

Gocnak (approved):
Absolutely fantastic work! Comparison:

Travis:
image

GitHub Actions:
image

That Server Test speed increase 🤤

PR #485, opened 2021-04-05
💄 Custom Loading Spinner

### LINKED ISSUE #443: Consider replacing generic loading spinner with a momentum logo animation

jedso:
This PR replaces the generic loading spinner with a "breathing" loading animation using the MMOD logo. First thing I tried was making the logo spin, but imo it didn't look great. 

This animation obviously looks better the longer it plays, but even on fast page load I think it looks fine. Here's how it looks with a throttled connection:

https://user-images.githubusercontent.com/26863777/113566231-72f05a00-964b-11eb-8654-85bf6a09d84b.mp4

Closes #443.


TheNerdMan:
I'm working atm so sorry if I'm asking too many questions, I can't pull repos on to this machine to build and take a look locally. 
 - When is load() in nebular's spinner.service called? Is that an us thing or something that nebular is handling? 
 - Is there an event we can hook into here?
 - Is it better to just build our own loading.service to be used anytime anything is loading?
 - Is nebular open-source/open to fixes?

Basically just using the links provided in this message I'd say first option would be to create a class that gets applied to the body around the hideSpinner() fire time. That then fades in the body (using opacity like you said). Another option is to try messing around with !important on some classes to override the display: none on the style attribute. Its not ideal but at the end of the day nebular really shouldn't be setting style attribute. However this still has the problem that we have no event to listen for when hideSpinner() is fired.

Otherwise I really wouldn't bother using this as like you said, display: none is an awful way to hide elements on the page, especially by setting the style attribute of them.

Hona:
Where is this at? @jedso @TheNerdMan

tsa96:
Dead PR and anims aren't great, closing.

PR #490, opened 2021-04-10
Fixed video text vertical centering on main page

### LINKED ISSUE #488: Gamemode video text is not vertically centered
image

vindow:
Fixes #488

First I centered the text and video using `align-items-center` but it left an annoying 4ish pixel gutter at the bottom of the video seen here:
image
so I opted to set its parent wrapper div as another row element so I could center it again using `align-items-center`. This 4 pixel gutter might be because of the YouTube embeds since it does not occur with those so when all the videos are WebM you might be able to remove this second row element workaround.

I also removed the bottom margin for the md breakpoint and up on the wrapper div since it would offset the video when the elements are stacked side-by-side.

Lastly I removed min-height from `gamemode-data-section` since it was causing an additionally large amount of margin padding  that wasn't present on the original version. It also did not affect the actual video size since the video is scaled based on width.

### DIFF
--- a/client/src/app/pages/main/gamemodes/gamemodes.component.html
+++ b/client/src/app/pages/main/gamemodes/gamemodes.component.html
@@ -38,8 +38,8 @@
       {{this.currentGameMode.modeTitle}}
     </div>
   </div>
-  <div class="row align-items-center mb-5 mb-lg-5 gamemode-data-section">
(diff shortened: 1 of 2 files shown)


Hona (approved):

Hona:
This looks like the better way to do it. Thanks for the work!

PR #497, opened 2021-04-26
Rewrite reports

murgle2:
Notes:

- Guid is non-nullable so resolverId defaults to "resolverId": "00000000-0000-0000-0000-000000000000",
- Pagination breaks because the list of objects counted becomes less with each offset.
             As a workaround, I get the total number of objects in the db and then limit by 5, which is equal to max results per page.
- Can't see summary strings on docs page
- Clicking "view reported resouce" doesn't take you to the profile page. Maybe missing functionality in user module?


API/src/Modules/Reports/Momentum.Reports.Core/Repositories/IReportRepository.cs
```diff
@@ -0,0 +1,16 @@
...
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Momentum.Reports.Core.Repositories
+{
+    public interface IReportRepository : IGenericRepository<Report>
+    {
+        Task<int> GetTodayReportCount(Guid submitterId);
+         Task<int> CountAllReports(bool resolved);
```
Hona, line 12: Spacing issues

API/src/Modules/Reports/Momentum.Reports.Infrastructure/MartenInitializer.cs
```diff
@@ -0,0 +1,14 @@
...
+using Momentum.Framework.Core.DependencyInjection;
+using Momentum.Reports.Core.Models;
+
+namespace Momentum.Reports.Infrastructure
+{
+    public class MartenInitializer : IMartenInitializer
+    {
+        public void SetupDomainIdentities(StoreOptions options)
+        {
+            options.Schema.For<Report>().Identity(x => x.Id);
```
Hona, line 11: In this case you don't need the line as Marten automatically matches the `Id` property, see [here](https://martendb.io/documentation/documents/identity/)

API/src/Modules/Reports/Momentum.Reports.Infrastructure/Repositories/ReportRepository.cs
```diff
@@ -0,0 +1,54 @@
...
+{
+    public class ReportRepository : GenericTimeTrackedRepository<Report>, IReportRepository
+    {
+        private readonly IDocumentStore _store;
+
+        public ReportRepository(IDocumentStore store) : base(store)
+        {
+            _store = store;
+        }
+
```
Hona, line 20: `GenericTimeTrackedRepository` provides this property: `.Store`, remove all this - and change `_store` usages to `Store`

API/src/Modules/Reports/Momentum.Reports.Infrastructure/Repositories/ReportRepository.cs
```diff
@@ -0,0 +1,54 @@
...
+        public async Task<int> CountAllReports(bool resolved)
+        {
+            using var session = _store.QuerySession();
+
+            var numberOfReports = await session.Query<Report>().CountAsync(x => x.Resolved == resolved);
+
+            return numberOfReports;
+        }
+
+        public async Task<IReadOnlyList<Report>> GetAllReports(int? limit, uint offset, bool resolved)
```
Hona, line 39: Change `int? limit` to `int limit = 5` (You'll have to make this the last parameter as optional parameters have to be last)

Hona (changes requested):

Hona:
> Guid is non-nullable so resolverId defaults to "resolverId": "00000000-0000-0000-0000-000000000000",

You can use the nullable Guid: `Guid?` instead.

API/src/Modules/Reports/Momentum.Reports.Infrastructure/Repositories/ReportRepository.cs
murgle2: If the limit parameter is changed to non-nullable, then it will default to 0. Then this default of 5 isn't applied and no reports are returned. For now I left this the same unless you have more suggestions

murgle2:
Using nullable Guid would require updating repo methods (in the report module as well as in the user module) with `Guid?` Not sure if this is worth doing?

PR #504, opened 2021-05-14
Added OpenCollective link and widget

### LINKED ISSUE #502: Add link to OpenCollective page on website
Momentum uses [OpenCollective](https://opencollective.com/momentum-mod) to accept donations. It would be good to have this page linked on the main page.

First thought is bare minimum: add it to the "Links" section, using the "gift" eva icon (`<nb-icon icon="gift">`), with the text "Donate"

Second thought is more involved and could come with a bit of a main page refresh (due soon with the engine update and branding update). It would be to make the OpenCollective link more prominent, like right below the "Play on Steam" embed in the main "header" section. Pending UX, and could actually be pulled out into its own card later.

vindow:
Closes #502

I added the donate link in the footer to the end of the list, not sure if that matters. 

As for the custom widget I did some digging on the official widget and [found how they drew their version.](https://opencollective.com/static/images/buttons/donate-button-blue.svg) I used that `<svg>` snippet to draw the logo and then created the rest of the button using CSS to basically look the same sans the color (there was also a barely noticeable gradient I didn't bother to add). I also copied how they spaced the text. 

With everything added the widget looks like this:
image

If the screen width is <380px (like on my dinky phone) it will look like this:
image

Finally, the footer looks like this:
image


client/src/app/pages/main/footer/footer.component.html
```diff
@@ -58,6 +58,12 @@
               <div>Docs</div>
             </a>
           </div>
+          <div class="mt-4 mx-1 link">
```
Gocnak, line 61: Yeah let's add this closer to the steam link near the left, it can be the second icon.

client/src/app/pages/main/header/header.component.html
```diff
@@ -7,6 +7,17 @@ <h2 class="text-white mx-auto mt-4 mb-5">A free, standalone movement-focused pla
         <div>
           <iframe src="https://store.steampowered.com/widget/669270/" style="border: none;" frameborder="0" width="70%" height="190"></iframe>
         </div>
+        <a class="donate" href="https://opencollective.com/momentum-mod" target="_blank">
+          <div class="donateBody">
+            <span class="donateText">DONATE TO OUR COLLECTIVE</span>
```
Gocnak, line 12: Let's change this wording to "SUPPORT US ON OPENCOLLECTIVE". Not sure how it'll affect the sizing or if this will properly size to fit.

Gocnak (changes requested):
And when you can, add some images to the original PR description with what the button looks like and the link in the links section.

client/src/app/pages/main/header/header.component.html
vindow: I had to widen the widget a bit to fit the text on one line. I also made it so if the screen width gets too small (<380px) it'll shorten the widget and put the text in 2 lines. Let me know if it looks fine like that.

vindow:
Also found this bug on small resolutions while reordering donate. `nb-icon` just wasn't listed in the css breakpoint at <768px in `footer.component.scss` so I just added it in to fix it.
image

client/src/app/pages/main/header/header.component.html
Gocnak: Yep it looks great!

Gocnak (approved):

PR #506, opened 2021-05-26
Added scrolling to global search results

### LINKED ISSUE #454: Global searches with enough results go visibly off-screen
I was unsure whether to make a new issue for this or to make a comment [here](https://github.com/momentum-mod/website/issues/263), but figured this would probably be better as the mentioned issue is vague in terms of bug fixing

image
image

In the first image, my browser is maximized; the second one my browser is in fullscreen
My display resolution is 1920x1080

vindow:
Fixes #454

The fix required setting the search results to have `overflow-y: auto` but in order for that to work you have to specify the height of the div.

Here's how it looks when you scroll down a bit:
image
If you want the Users and Maps headers to stay visible it would have to be broken down into two scrollable columns or the HTML layout would be need to be reworked so that the headers are in their own separate div.

### Comparisons
I set the top margin to `10vh` instead of `10%` to simplify the height calculation for `results`. This change makes it look slightly different from before. New on the left, old on the right.
Here's how they look on an average 1920x1080 screen (non-fullscreened browser):
image

And here's how they would roughly look on a phone (phone resolutions differ between phones):
image

### Height Calculation
The `calc` statement for the results div height is a little jank. Here's the breakdown:

- `90vh` The top margin is now 10vh so we start with 90vh left
- `8vw` The input text is scaled as `6vw` but it seems there's a bit of extra space allocated (maybe line spacing?). image 
Regardless, I just trial and errored my way to `8vw`
- `5rem` Calculated based on: `0.9`(hint text size) + `0.85 * 2`(padding top and bottom of hint text) + `0.25 * 2` (padding on input) + `1.5` (top margin of results).
- `4px` The input underline is a 4px bottom border.

I also added 0.4rem of padding on the bottom so when you bottom out the search results it doesn't look weird.

All of this seems to work. I tested extremely skewed resolutions (4000x500, 400,2500)  along with more standard resolutions and it seems to be aligned properly.

### Scrollbars
I hid the scrollbar for this behavior since I didn't think it looked good:
image
Note that I had to use `scrollbar-width: none;` for FireFox and `::-webkit-scrollbar` for Chrome since apparently they don't both do it the same way.
I'm also not sure where to disable main page scrolling when the search modal is open if that's a desired behavior. Applying `overflow: none` to the various parent/body elements when I was experimenting in the browser didn't seem to do anything so I'm not sure where to even begin looking in that case.

### DIFF
--- a/client/src/app/@theme/components/search/search-field.component.html
+++ b/client/src/app/@theme/components/search/search-field.component.html
@@ -14,7 +14,7 @@
       </div>
       <span class="info">{{ hint }}</span>
       <div class="mx-auto mt-4 content">
-        <div class="row justify-content-center">
+        <div class="row justify-content-center results">
           <div class="col-10 col-md-6" *ngIf="users?.length">
             <search-results [user]="true" [onlyMaps]="onlyMaps" [onlyUsers]="onlyUsers" [elems]="users"
                             (selectedURL)="selectedElem($event)"></search-results>
--- a/client/src/app/@theme/components/search/search-field.component.scss
+++ b/client/src/app/@theme/components/search/search-field.component.scss
@@ -1,6 +1,6 @@
 .results {
   overflow-y: auto;
-  height: calc(77vh - 5.4rem);
+  height: calc(90vh - 8vw - 5rem - 4px);
   scrollbar-width: none;
 }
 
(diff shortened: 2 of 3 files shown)


Gocnak (approved):
Looks great! Thank you for also taking the time to show before/after photos in the description, really helps visualize the changes here!

PR #519, opened 2021-07-05
Migrate to ESLint

bakapear:
We should migrate to ESLint since TSLint is deprecated.

However ESLint comes with its new rules I had to temporarily disable in 54eeb6f to pass through the lint check. Most of these rules should be enforced though.


Gocnak:
To clarify:
> Most of these rules should be enforced

Does this mean we get those rules for free? I'm confused as to why they're explicitly turned off. Are you aiming to turn them back on in future PRs? Should we make a github issue about re-enabling them?

bakapear:
I turned them off to mirror the current coding style, so this PR can be safely merged without any change in linting.
But yes, we would have to discuss about the new ESLint styles in a new issue and lint the entire project with the new rules in later PRs.

client/package.json
```diff
@@ -76,14 +76,25 @@
...
     "@angular/compiler-cli": "9.1.11",
     "@angular/language-service": "9.1.11",
     "@fortawesome/fontawesome-free": "5.13.1",
     "@types/jasmine": "3.5.11",
     "@types/jasminewd2": "2.0.8",
     "@types/node": "^12.12.30",
+    "@typescript-eslint/eslint-plugin": "4.3.0",
+    "@typescript-eslint/parser": "4.3.0",
     "codelyzer": "5.2.1",
+    "eslint": "^7.6.0",
```
Gocnak, line 94: We use specific versions since we're not a library, so this caret can be removed in favor for the latest stable 7.6.X version number here that satisfies the other eslint dependencies added.

Gocnak (approved):

PR #521, opened 2021-07-10
Update frameworks

bakapear:
Angular 9 -> Angular 12
Nebular 5 -> Nebular 8
and other packages that needed updating to support the new versions.

### Notes:
- Nebular throws a Sass division deprecation warning, which can be fixed by following the [automatic migration](https://sass-lang.com/documentation/breaking-changes/slash-div#automatic-migration), but it has to be done by the nebular package maintainers themselves.
- I did not update to Bootstrap 5, as it breaks the entire site's layout.
- Angular 12 dropped support for IE 11 and lower. These have been removed from the browserslist.


Hona (approved):

Gocnak (approved):

Hona:
Checked the client artifact, looks like nothing broke 👍

PR #524, opened 2021-07-26
Use SVG for the Momentum Mod Logo

bakapear:
Replaces the logo with a SVG version found in the game files.

The difference is subtle but it basically removes the blurriness when scaled down.

| Before & After |
| - |
image
image


Gocnak (changes requested):
Let's also remove the old PNG file.

Gocnak (approved):

PR #525, opened 2021-07-30
Main page BG branding update / Branding implementation

### LINKED ISSUE #492: Link Press Kit Page to Bottom of Main Page
we need a page where anybody writing a blog/article/etc can easily go to quickly figure out what our game is about and access some needed resources. should be somewhere fairly easy to access on the main page and would be relatively simple. 

at a minimum it should include:
1. a bit about us and what we do, mission statement stuff
2. some statistics people can easily reference (like game downloads? idk)
3. (most importantly) easily downloadable high quality versions of logos, backgrounds, other assets.

some examples of existing press kits that we can reference are:
- https://www.annmariegianni.com/press-kit/
- https://about.twitter.com/en/who-we-are/brand-toolkit

### LINKED ISSUE #523: Update main page BG branding to match new brand direction
Out with the grey icons, in with the light and lively 3D crystalline geometry!

Using a video tag that is streamed in (use `preload=metadata`), and use the photo as the "banner":
[MomentumLight.zip](https://github.com/momentum-mod/website/files/6875708/MomentumLight.zip)

See the gamemode component for how it was done there:
https://github.com/momentum-mod/website/blob/staging/client/src/app/pages/main/gamemodes/gamemodes.component.html#L45

comatoad:
Closes #523 and closes #492 

Posterity:

- Updated frontpage to use .webm (4mb) on non-mobile devices. Casts a gradient over for a similar shadow effect using sass, implementation took a bit of tweaking (going above a width of 99.1vw will probably cause things to explode)
- If on a mobile device, the .webm is not loaded and the .webp is loaded (~370kb). Should be faster and fixes a very annoying gradient error 
- 404 page displays .webp
- 'Branding' added at the bottom linking to the pdf inserted in /assets/
- Navbar has significantly less transparency than prior; this can (and probably should be) tweaked. I tweaked the momentum black colors' transparency until it looked decent, but UI/UX might have a better idea


client/src/app/pages/main/header/header.component.html
```diff
@@ -1,5 +1,9 @@
 <section id="header">
+  
   <header class="masthead">
+    <div class="gradient">
+      <video class="banner-video" src="../../../../assets/images/bg-main.webm" poster="../../../../assets/images/bg-main.webp" loop autoplay ></video>
```
Gocnak, line 5: Is this properly pre-loading? As in, the website loads fully and this asynchronously loads and then replaces the "poster" when it does?

comatoad: I believe so, but I actually think this could be axed. If a poster tag isn't included I believe it uses the first frame of <video> as the poster instead, and since it's the same source material we could optimize it further by leaving it out

Gocnak: I think it's fine to keep the poster. I read about preload some more and in our use case, with `autoplay`, we'll want it to load immediately anyways. So this should be fine!

Gocnak (approved):
Very nicely done! Can't wait to see what it looks like on the main site. 👍

PR #527, opened 2021-08-01
Feat/fixbg

### LINKED ISSUE #526: Dashboard background is not removed when navigating back to main page
image

Looks like if you access the dashboard and navigate back to the main page, the background is cached and overlaps with the video. This is due to a pseudo ::before element (background). Probably just needs to update the background tag (null it out) or use a different approach - will r&d

Edit: might be hard to test, I assume adding `background: none` to header.component.ts should fix the issue, but I cannot recreate it locally as the bug uses the dashboard - see relevant [commit](https://github.com/comatoad/website/commit/8e462de82e97b4e90b359be9cffb5a335e8f2b3d)

wallenben:
Posterity:
Fixes issues mentioned in #523
Closes #526

- autoplay should work better (thank you @Nicell ;))
- Gradient displays properly and fades in properly (thank you @Nicell)
- Removed dashboard background in the immediate to prevent crazy pseudo-class errors
- Extraneous sizing and aspect ratios should display properly


client/src/app/@theme/layouts/sample/sample.layout.scss
```diff
@@ -43,7 +43,7 @@
     height: 100vh;
     z-index: -1;
     will-change: transform;
-    background: url('/assets/images/logoBG.png') no-repeat center;
+    //background: url('/assets/images/logoBG.png') no-repeat center;
```
Gocnak, line 46: Because we're on source control, these lines can be outright removed.

If this is the only use of this `logoBG.png` file, you may remove it as well.

client/src/app/pages/main/header/header.component.scss
```diff
@@ -5,10 +5,10 @@
   display: flex;
   width: 100%;
   height: 100vh;
-  min-height: 35rem;
-  padding: 15rem 0;
+  min-height: 55rem;
   z-index: 0;
   overflow: hidden;
+  background: rgba(22, 22, 22);
```
Gocnak, line 11: Does an `rgba` without an alpha value default to 1.0? If so, could this be hexcode instead?

Gocnak (changes requested):

client/src/app/@theme/layouts/sample/sample.layout.scss
wallenben: Sounds good. I have removed the line and the file.

client/src/app/pages/main/header/header.component.scss
wallenben: Yes it does! I'll change it to hex and push the commit.

Gocnak (approved):

PR #534, opened 2021-08-07
Feat/enterkey

### LINKED ISSUE #533: Filtering the "browse maps" page doesn't accept the enter key
At `https://momentum-mod.org/dashboard/maps` typing something into the box to search for a map and then hitting enter does nothing. You have to mouse over to the 'search' button instead which is unintuitive and should be changed.

image

Also worth noting is the fact that selecting a gamemode from the dropdown next to the search box also requires you to hit the 'search' button to update your results. Maybe this should automatically apply when selecting a different gamemode instead. 

image

wallenben:
Closes #533 

Pressing the enter key should automatically refresh the map list now. Also, updating the "game mode" list should instantly trigger a map list refresh. This should be tested by someone with API access prior to double-check that the refreshes are functioning properly, just in case.

Edit: Accidentally based this branch off features for another issue. Should be reverted now.


client/src/app/pages/dashboard/maps/map-list/map-list.component.html
```diff
@@ -3,7 +3,8 @@
     <div class="col-12 col-sm-9 col-md-10">
       <div class="row align-items-center">
         <div class="form-col col-12 col-sm-10 col-md-6 col-lg-4 col-xl-3">
-          <input nbInput fullWidth class="form-control" type="text" placeholder="Map Name" formControlName="search">
+          <input nbInput fullWidth class="form-control" type="text" placeholder="Map Name" formControlName="search" onkeydown = "if (event.keyCode == 13)
+          document.getElementById('searchBtn').click()">
```
Gocnak, line 7: Noting here, as mentioned on the call, this would probably be better if it were a form, where pressing enter submits the form (applies the search).

See [the guide on forms](https://angular.io/start/start-forms) for more info.

Gocnak (changes requested):

client/src/app/pages/dashboard/maps/map-list/map-list.component.html
wallenben: Sounds good. I have discarded that change and added a form in its place - doing a little extra research to see if I can do something similar for the dropdown to keep the code clean.

wallenben: Update on the dropdown: while this solution works for now, I talked to TheNerdMan about potentially refactoring the button down the line to be more efficient with API usage and overall Angular principles using two-way bindings. If it's alright with you, I think it could be worth checking out as a new card enhancement; something I can tackle later as I learn how to tackle this problem

wallenben:
closed in favor of nerdman's upcoming optimized implementation

PR #538, opened 2021-08-16
💄 Error & Maintenance Pages

### LINKED ISSUE #356: Custom maintenance page
When the server is down for maintenance, create an HTML page styled to be simplistic but better than the 502 Gateway Error found with nginx.

Relevant links:
https://lincolnloop.com/blog/pro-tip-redirecting-custom-nginx-maintenance-page/

### LINKED ISSUE #528: 404 page has poor visibility with new background image
image

See https://momentum-mod.org/asdfsd

wallenben:
Closes #356 
Closes #528. 
This updates the 404 page to have the text be visible (yay)! Preview shown below:
image
I tried to make it as 1:1 with the render as possible - the black bar definitely helps quite a bit here.

As for the 502 page, I recreated a 'lite' version of the 404 page - because there is no button component needed, some tweaking was done. I opted to not include the gradient as I felt we wanted this page to be pretty light as-is, but it's a very quick implementation if needed. Stylistically this file fought me quite a bit (certain CSS properties didn't want to inherent correctly) so any suggestions for improvement here would be appreciated
image


Gocnak:
Could you squash 4e18d55 into e130a3e ?

bakapear:
### Issues:
<details>
<summary>404 - Background size not consistent</summary>

(Also the scrollbar shows all the time)
Hectorsdolphin_35318+ x +(2021-08-17+at+09 21 58)
</details>
<details>
<summary>502 - Not responsive at all</summary>

image
</details>
<details>
<summary>502 - Custom font not loading</summary>

Gotta add the `Bebas Neue` font to the maintenance folder too, I think.
</details>

### Suggestions:
<details>
<summary>502 - More user-friendly title message</summary>

(Add `Error Code: 404` to 404 page aswell then)
image
</details>
<details>
<summary>502 - Optional Discord button</summary>

Might be too much for a simple 502 but idk
image
</details>
<details>
<summary>Background gradient</summary>

I do prefer the non-gradient look on the server message pages but it IS a little to bright without it in my opinion (probably just me because I have everything on windows set to darkmode). Maybe change the gradient to be a consistent darker overlay.
image
</details>

wallenben:
Thanks for all the tips! I'm gonna be a bit AFK (moving time), but I will crack away at these suggestions. Pretty sure I can pull the Angular breakpoints for super easy 502 reactiveness. The background sizing error is 100% due to me forgetting a CSS property so I can spin that up super quick as well.

I like the 'page under maintenance' as the header waaaay better, users probably don't know/care about error codes. I'll take a look into the discord button - I think if I chop up some of the main site's CSS it shouldn't be too tricky but it may cause sizing issues. I also feel the gradient helps tone it down, so I'll look into that

wallenben:
One more thing - I looked into removing the scrollbar from the 404 page, but I believe it's due to the use of `<nb-layout>`. This is supported by the issue not being present on the 502.html document as it can't use that property. Is this something we would want to tackle in the future?

bakapear:
Good idea to reduce the letter spacing on the header. It's quite long.

I think there is still some tweaking with the background you could still do:
<details>
<summary>The background height shouldn't change size here</summary>

Preview
</details>
<details>
<summary>It would also probably be better if the background size is changed with the width breakpoints</summary>

Preview
</details>

I have a commit here 3373c86f9d474d5f12a52151b2860443bb5d32f1 which resolves both of these (by just changing CSS that adds a background color to the text div block instead of adding a new html tag for the background itself).

The top and bottom gradients look a bit too dark for me now but otherwise it's fine!

As for the scrollbar issue- it occurs on every page. That would need to be fixed in a separate PR as it's not related to only error pages.

wallenben:
Back from the hiatus!
In one of the team meetings, it was discussed that switching this to the black background would be aesthetically superior, so I will be reformatting both versions to work with this. I'll put a preview below when I have something ideal.

I'll add that commit that fixes those nasty issues as well - thanks for that!

Hona:
How are you going with this @wallenben

wallenben:
I added bakapear's fixes to this which solved the aforementioned sizing issues neatly. For posterity, the added commits make the pages look like so:
chrome_uuf6ug6l2e

chrome_Kjdky2q2ub

Couldn't find much on the scrollbar thing, but I'll keep looking to see what can be done.

Gocnak (approved):
The scroll issue was fixed in #566 so this should be good!

PR #557, opened 2021-11-05
👷‍♂️ Use docker-compose integration tests

murgle2:


.github/workflows/tests.yml
```diff
@@ -0,0 +1,17 @@
+name: Tests
+
+on:
+  pull_request:
+    branches:
+      - staging
+    paths-ignore:
+      - '**.md'
```
Hona, line 8: Shouldn't this also run on push to staging?

Hona (changes requested):

Hona:
I think you should also explicitly choose the yml files with `-f docker-compose.yml -f docker-compose.override.yml`

.github/workflows/tests.yml
murgle2: Added

Hona (approved):

PR #561, opened 2021-11-19
👷 Fix server tests

murgle2:


Hona:
I would suggesting changing this title to 'Fix server tests' (or similar) and include all fixes, like missing `.env`, the port, missing db script etc

@murgle2

PR #559, opened 2021-11-09
🐳 Add docker-compose

murgle2:

PR #562, opened 2021-11-19
👷 Fix docker-compose integration tests

murgle2:


Hona:
Please resolve the merge conflicts.

Also, it looks like the github actions aren't running on the latest commit here?

murgle2:
> Please resolve the merge conflicts.
> 
> Also, it looks like the github actions aren't running on the latest commit here?

Resolved!

Hona:
Is this ready for test?

murgle2:
> 

Yep! It just needs that other PR (#561) so it will test the docker part correctly

Gocnak:
Server tests aren't running right now. It's building the containers fine but the output is pretty much meaningless (if it's actually running tests, which considering GitHub actions moves on immediately, I don't think so):
image

Gocnak (changes requested):
Server tests aren't running right now. It's building the containers fine but the output is pretty much meaningless (if it's actually running tests, which considering GitHub actions moves on immediately, I don't think so):
image

I would also prefer for the Docker containers to use a Yarn cache so we aren't waiting minutes again. 
image

This has been an issue since #559, I'm not sure what the rush was to get it in. @TheNerdMan also pointed out that there's no frontend dev (with auto-reloading) container either, which significantly hinders development progress.

murgle2:
Server tests now shows tests outputs and correctly uses the exit code to determine test results. 

The benefit of dockerizing the backend is very significant in comparison to the frontend. The current dev workflow involves running yarn install and npm start for the frontend, which does autoreloading and communicates with the dockerized backend, which also autoreloads. 

If there's a strong demand for it, I can dockerize the frontend as well.

server/package.json
```diff
@@ -4,8 +4,9 @@
   "private": true,
   "scripts": {
     "start": "node app.js",
+    "dev": "nodemon server.ts -L --ext ts,json,js",
     "ts-build": "tsc",
-    "test": "mocha --exit --timeout 45000"
+    "test": "chmod +x ../scripts/wait-for-it.sh && ../scripts/wait-for-it.sh db:3306 --timeout=60 -- mocha --exit --timeout 45000"
```
Gocnak, line 9: I think the `"test"` command should still be the `mocha --exit --timeout` stuff, and this specifically should be `test-docker`:

```suggestion
    "test": "mocha --exit --timeout 45000",
    "test-docker": "chmod +x ../scripts/wait-for-it.sh && ../scripts/wait-for-it.sh db:3306 --timeout=60 -- npm run test"
```

That way we're separating out the actual test command from the docker entrypoint command.

docker-compose.test.yml
```diff
@@ -19,4 +19,4 @@ services:
     build:
       context: ./server
       dockerfile: dev.Dockerfile
-    entrypoint: sh -c "chmod +x ../scripts/wait-for-it.sh && ../scripts/wait-for-it.sh db:3306 --timeout=60 -- npm test"
+    entrypoint: "npm run test"
```
Gocnak, line 22: And with the other suggestion:
```suggestion
    entrypoint: "npm run test-docker"
```

Gocnak (changes requested):
Yeah the frontend just being a local dev environment is fine! The backend was the more tedious setup anyways, so great that it's properly containerized now.

I'd say caching is absolutely still worth it but can be done in another branch. A few nits here but I'm otherwise fine to merge this in after.

Gocnak (approved):

PR #563, opened 2021-11-29
⚡ Increase request limits for certain map data

ijre:
As Goc had mentioned and further suggested\*, this PR is for increasing the query limit for map info and the user's own library info from 20 to 100

\*suggested in the sense that he had said 100 requests and I figured I'd go with that

I'm pretty sure that there's no other places defined limiting this specific data, but I'm not 100% tho I did look

### DIFF
--- a/server/src/models/map-library.js
+++ b/server/src/models/map-library.js
@@ -10,7 +10,7 @@ module.exports = {
 		const queryOptions = {
 			distinct: true,
 			where: {userID: userID},
-			limit: 20,
+			limit: 100,
 			order: [['createdAt', 'DESC']],
 			include: [
 				{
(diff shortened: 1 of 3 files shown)


Gocnak:
Needs rebasing onto latest staging

server/src/models/map-library.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
 		const queryOptions = {
 			distinct: true,
 			where: {userID: userID},
-			limit: 20,
+			limit: 100,
```
Gocnak, line 13: If we're updating map-library, map-favorite should probably also be updated, too.

Gocnak (changes requested):

Gocnak (approved):

PR #571, opened 2021-12-16
♻️ Update Docker Compose setup

murgle2:
- Move env var for test to the compose file so no longer need to update .env file
- Clean up read me
- Use project tag on test docker exec so doesn't overwrite dev services
- Simplify compose setup
- Add host property to env.TEMPLATE for prod env


server/README.md
```diff
@@ -28,26 +28,13 @@ Navigate to `http://localhost:3002/api-docs` to view the Swagger server API refe
...
 
-Copy the env.TEMPLATE file to .env and set this variable:
-```
-NODE_ENV=test
-```
-
 From the website directory, run:
 ```
-docker compose -f docker-compose.yml -f docker-compose.test.yml up -d
+docker compose -f docker-compose.yml -f docker-compose.test.yml -p test up -d
```
Gocnak, line 33: Should the workflow also be updated to run this command? What does `-p test` do?

docker-compose.prod.yml
```diff
@@ -1,17 +0,0 @@
...
-      MYSQL_USER: ${MOM_DATABASE_USER}
-      MYSQL_PASSWORD: ${MOM_DATABASE_PW}
-      MYSQL_ROOT_PASSWORD: ${MYSQL_ROOT_PW}
-
-  api:
-    image: docker.pkg.github.com/momentum-mod/website/mmod-website:staging
-    volumes:
-      - ./server/public/img/maps:/app/server/public/img/maps
-      - ./server/public/maps:/app/server/public/maps
-      - ./server/public/runs:/app/server/public/runs
```
Gocnak, line 17: We don't need to delete this. I want to move production over to dislocated docker images like this, helps separate the responsibility and how much the VPS has to set up.

I'd re-add it but change the image: to be `ghcr.io/momentum-mod/website/mmod-website:staging` instead (using GH'd new URLs)

Gocnak: Similarly, NODE_PORT needs defined, as production uses that in config.js

server/README.md
murgle2: -p test gives a project label to the containers. This gives them a unique name so that the test containers and the dev containers won't overwrite one another.
image

This is only necessary in dev as the workflow won't be running the dev containers, only the test containers.

Gocnak:
My work over on #575 will conflict with this branch.

Hona:
@murgle2 What is the plan for this now that #575 has been merged?

murgle2:
> 

Merged the changes. I think this is still useful as it simplifies running unit tests locally

tsa96:
Closing, rewrite covers all this besides docker prod, which me and Goc will work on in the coming month or so

PR #572, opened 2021-12-17
Add "STEAM_PREVENT_LIMITED" env var + fixup tests workflow file

Gocnak:
Checking STEAM_PREVENT_LIMITED, if it's "true" then it'll prevent limited accounts from connecting

This logic will be refactored to only "disable" limited accounts by flagging them and preventing uploading runs in the future

Updated workflow file to:
1. Only run the appropriate jobs when it detects files to run them against have changed
2. Actually run lint again (it caused some problems with the recent merge...)
3. Only run on pull requests, as direct pushes to staging did not need to be tested again

### DIFF
--- a/.github/workflows/angularbuild.yml
+++ b/.github/workflows/angularbuild.yml
@@ -1,7 +1,5 @@
 name: AngularBuild 
 on:
-  # Trigger the workflow on pull request,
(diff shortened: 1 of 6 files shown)


TheNerdMan (approved):
Looks good!

PR #575, opened 2021-12-22
Add S3-Compatible Cloud Storage to Maps, Images, and Runs

### LINKED ISSUE #316: Store runs, map images, and map files in cloud storage
Currently the best bet is going to be [Backblaze](https://backblaze.com) as they can be used [with our Cloudflare free plan](https://help.backblaze.com/hc/en-us/articles/217666928-Using-Backblaze-B2-with-the-Cloudflare-CDN) to have free CDN distribution of our files.

But if we have purists:
AWS S3 is $3/mo for 100GB and 100k requests for GET/PUT, respectively.

[Here's a sample page](https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/getting-started-nodejs.html) that does S3 uploading using AWS S3 for Node.

Gocnak:
Closes #316 

For development we use MinIO in docker compose to act as our S3-compatible platform. Production will use Backblaze.

I did refactor a bit of the local storage so this can't be deployed until we're fully ready to switch over to cloud storage anyways.


murgle2:
Looks good! Maybe some updates in the readme regarding minio and I think this is good to go

murgle2 (approved):

server/src/helpers/filestore-cloud.js
```diff
@@ -0,0 +1,45 @@
...
+	forcePathStyle: true,
+	signatureVersion: 'v4',
+});
+
+module.exports = {
+
+	storeFileCloud: async (fileBuffer, fileName) => {
+		const results = await s3Client.send( new PutObjectCommand( {
+			Bucket: config.storage.bucketName,
+			Key: fileName,
```
TheNerdMan, line 19: This should be called file key, if we want to store more things in cloud then we'll be separating them out using a folder structure-like key for it (E.g "maps/filename.extension" or "asset/atlas/filename.extension"). It's just a small thing but naming the functions args will clear that up

server/src/helpers/filestore-cloud.js
```diff
@@ -0,0 +1,45 @@
...
+
+		const hash = fsutil.getBufferHash(fileBuffer);
+		return {
+			fileName: fileName,
+			downloadURL: downloadURL,
+			hash: hash,
+		}
+	},
+
+	deleteFileCloud: async (fileName) => {
```
TheNerdMan, line 36: Same here, fileName -> fileKey

server/src/helpers/filestore-local.js
```diff
@@ -0,0 +1,96 @@
+'use strict';
+
+const config = require('../../config/config'),
+	 fs = require('fs'),
+	 fsutil = require('./filestore-utils');
+
+module.exports = {
+	// fileName is most likely 'maps/mapID'
+	storeMapFileLocal: async (mapFileBuffer, fileName) => {
```
TheNerdMan, line 9: Same as comment on server/src/helpers/filestore-cloud.js : line 19

TheNerdMan (commented):
See comments, they are nit picky but apart from that everything looks good.

Are we going to worry about checking the hash of the file that is uploaded to ensure there was no corruption during upload? Small thing but just more failure checking

server/src/helpers/filestore-local.js
Gocnak: Not sure about it for local since it is quite literally writing an actual file named that.

server/src/helpers/filestore-cloud.js
Gocnak: Changed in latest push.

server/src/helpers/filestore-cloud.js
Gocnak: Ditto, fixed in latest.

server/src/helpers/filestore-local.js
Gocnak: Similarly, we're going to be removing local storage in the nest rewrite, this is only backwards compat for now.

TheNerdMan (approved):
😍😍😍😍

PR #580, opened 2021-12-31
✨ Delete Runs

murgle2:
Allow admins to delete a users run and shift the other runs to fill in the rank gap. Checks if the user has a next best run and inserts that run.

Notes:
- Does not reset stats gained once the run is deleted
- If the next best run replay does not exist on server, the replay will only be accessible locally
- If a user has tied times as their next best run (same total ticks), then the next best run may not be the most recent
- Since tickrate datatype was changed to decimal this will require db sync

### DIFF
--- a/server/src/models/run.js
+++ b/server/src/models/run.js
@@ -645,57 +645,90 @@ const updateStats = (resultObj, transaction) => {
 const updateRanksAfterDelete = (oldRank, nextBestRun) => {
 
 	const deletedRank = oldRank.rank;
-
 	let rankSearch = {[Op.gte]: deletedRank};
+	let completions = 0;
 
 	return sequelize.transaction(transact => {
-		return UserMapRank.findAll({
+		// get total completions
+		return UserMapRank.count({
 			where: {
 				mapID: oldRank.mapID,
 				gameType: oldRank.gameType,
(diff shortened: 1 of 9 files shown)


server/src/models/run.js
```diff
@@ -642,6 +642,98 @@ const updateStats = (resultObj, transaction) => {
...
+				for (const result of results) {
+					result.rank -= 1;
+					result.rankXP = xpSystems.getRankXPForRank(result.rank, completions).rankXP;
+					updates.push(result.save({transaction: transact}));
+				}
+				return Promise.all(updates);
+			});
+		}).then(() => {
+			// if we have a nextBestRun, we need to update oldRank
+			if (nextBestRun) {
```
Hona, line 686: Make this a guard statement

Hona (changes requested):
Looks pretty good!

tsa96:
Closing, as this is way too outdated to ever be useful. I really appreciate the work done here and sorry that this can't be used. It's very frustrating how disorganized the website work was in this time period, pretty sure we knew the backend was going to need rewriting at this point already, and shouldn't have encouraged people work on issues like this.

tsa96:
Also, I will never accept a PR containing a Gitmoji

PR #611, opened 2022-03-17
🎨 - Remove trailing commas from Prettier

tsa96:
I have no idea why Prettier changed to use 'es5' as default, but for TypeScript it's absolutely horrible, it errors for trailing commas on type params. I don't think anyone's a fan of trailing comma, let's remove. Should be the last thing, after which we'll hopefully never have to touch Prettier config again.

### DIFF
--- a/server/.prettierrc.js
+++ b/server/.prettierrc.js
@@ -1,6 +1,6 @@
 module.exports = {
     printWidth: 120,
-    trailingComma: 'es5',
+    trailingComma: 'none',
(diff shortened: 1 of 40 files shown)


TheNerdMan:
Now that I think about it, trailing commas look like someone is still developing and stopped half way. A welcome change in my eye's. I'm assuming this has been built and checked it doesn't break any prisma stuff?

TheNerdMan (approved):
I Love it

PR #612, opened 2022-03-19
🐛✨ - Logger Fixes and Enhancements

TheNerdMan:
The NestJS logger will be mostly what we need, for now. However we needed some extra steps for logging objects (just using `JSON.stringify` for now).

I've also added a HTTP middleware logger that will log all incoming requests (Ignores the `/auth` endpoints though, this my contain sensitive info)


tsa96 (approved):
Great stuff, the HTTP middleware is really nice.

PR #614, opened 2022-03-22
✨ - Error handling exception filtering

tsa96:
Adds a exception filter to main.ts, which appends error code and message to client whenever an exception is hit. Sends a generic error for database errors. Logs errors with same error code for easy tracing.


TheNerdMan (approved):
Hey I'm in this!

PR #615, opened 2022-03-23
🩹 - Rename exception filter

tsa96:
I'm an idiot. Forget to update the filter name, we're catching everything with this now. Also, the ternary was the wrong way round, forget to revert back after changing to test stuff.

In the future if we're both working on stuff together and it's late, let's review the following day with fresh eyes.


TheNerdMan:
> In the future if we're both working on stuff together and it's late, let's review the following day with fresh eyes.
Couldn't agree more, I was way to tired to properly review that. Won't be rushing it in the future

TheNerdMan (approved):
Makes a lot more sense now

Issue #536, opened 2021-08-14
Optimizing the map search component further with two-way bindings


wallenben:
Extension of discussion on #534 
Currently, map data is handled through a form. Refactoring the map form and the game-mode dropdown to be more efficient with API usage is possible, so it's definitely worth looking into.

![image](https://user-images.githubusercontent.com/72104872/129456064-3225eb47-b76d-41d3-8db6-ed82447175c3.png)

Relevant discussion points from TheNerdMan:

> I'd set up two way bindings on the text field, and dropdown. Have a "onChange" event that fires when text changes,drop down changes or button is clicked. Then that "onChange" event will check those twoway binding values and makes the appropriate API call. 

> [You should] hook up the two-way bindings to the state's properties, that way if there are other map filtering places, those search elements will be saved and reused

Further examples are posted in the webdev channel in discord.

tsa96:
Bit confused by this issue. I agree with Alex that we should use a two-way binding here instead of a form, so users don't have to press enter or a "search" button - I'll do that in my incoming PR.

This isn't a performance optimisation though, we'll be making more queries. But I don't think the performance is very in important here as the queries are relatively cheap, since
- this list doesn't need to fetch a ton of data about each map, really just the primary table and the thumbnail, so only one join
- the MMap table is relatively small, but queried a lot. so it can have a whole bunch of indexes on at minimal costs
- the frontend doesn't get used that much

So my incoming PR will essentially implement this and close this issue.

tsa96:
Okay, incoming PR will close this. Two way binding isn't needed, Angular forms are perfectly capable of handling this - just use a subscription to the `FormGroup.valueChanges` observable.

PR #621, opened 2022-04-06
Combine favorites/library into browse

### LINKED ISSUE #602: Simplify maps lists, add filtering
**Is your feature request related to a problem? Please describe.**
The "My Library", "Favourite Maps" and "Browse Maps" are 3 pages of lists doing essentially the same thing.

**Describe the solution you'd like**
Remove the "My Library" and "Favourite Maps" pages. Make the library and favourites buttons on the "Browse Maps" pages more prominent, with a tooltip somewhere explaining what adding a map to your library actually does i.e. makes the game download it for you automatically. Add filtering based on your favourites and library similar to the in-game map selector UI.

**Additional context**
Had brief discussion with Gocnak about this in #website on 25/02/22

vindow:
Closes #602

Since it was fairly trivial I also threw #533 in this but the form will need to be tweaked for #536 so there's still work to be done later. I'm not sure if you should close that one and leave #536 open. The dropdowns for status and type are also updated to submit the form when changed.

I used nebular toggle buttons in the filter form but they might be a little hard to see when untoggled. New is on the bottom:
fullwidth
The favorite and library buttons were also slightly enlarged.

Here's what the tooltip for library looks like:
image

I also changed the responsive styling of the filter form to just use flex box, the map name input shouldn't resize anymore and is a fixed width. Here's how it looks on iPhone SE:
image
The two new toggles move to a new row together when flex wraps since they're in the same div.

Lastly, I removed the navigation links and the routing for the old pages and deleted their components.
image


TheNerdMan (approved):
Everything looks in place and after running and testing it out it seems fab.

Love this, love the code xox

PR #623, opened 2022-04-15
fix: Exception filter changed all errors to 500

TheNerdMan:
Found this when I should have got an unauthorized error but instead was returned a 500. A quick spruse up of Tom's exsisting work and we have something that fits all HTTP errors a bit better while retaining the functionality of hiding the details of DB errors from clients and having easily lookup able logs with the error codes.


tsa96 (approved):
Good, the separate http error class is nice.

PR #626, opened 2022-04-17
feat: Error reporting + Performance monorting

TheNerdMan:
- Sentry implemented
- Removed old health + metrics endpoints
- Removed Grafana and promethus UI's from docker compose

You'll need the DSN from Sentry so login to get that.


server/src/app.module.ts
```diff
@@ -1,29 +1,58 @@
...
 export class AppModule implements NestModule {
     public configure(consumer: MiddlewareConsumer): void {
         consumer
+            // Accept raw posts from these paths
             .apply(RawBodyMiddleware)
             .forRoutes({
                 path: '/auth/steam/user',
                 method: RequestMethod.POST
             })
+            // Accept JOSN from these paths
```
Gocnak, line 45:
```suggestion
            // Accept JSON from these paths
```

env.TEMPLATE
```diff
@@ -28,6 +28,10 @@ API_URL=
 AUTH_URL=
 CDN_URL=
 
+# Sentry details for logging
+SENTRY_DNS=
+SENTRY_DEBUG=
```
Gocnak, line 33: What does SENTRY_DEBUG do? What are the expected values?

Gocnak (changes requested):

env.TEMPLATE
TheNerdMan: SENTRY_DEBUG just logs stuff to console when creating and ending the transactions. Expects true/false

env.TEMPLATE
```diff
@@ -28,6 +28,10 @@ API_URL=
 AUTH_URL=
 CDN_URL=
 
+# Sentry details for logging
+SENTRY_DNS=
```
Gocnak, line 32:
```suggestion
SENTRY_DSN=
```

Gocnak (changes requested):

TheNerdMan:
Reminder for me: If no DSN is set then don't init Sentry as this causes issues

server/src/@common/sentry/sentry.module.ts
```diff
@@ -0,0 +1,42 @@
...
+            providers: [
+                {
+                    provide: SENTRY_OPTIONS,
+                    useValue: options
+                },
+                SentryPerformanceService,
+                SentryExceptionService,
+                {
+                    provide: APP_INTERCEPTOR,
+                    useClass: SentryInterceptor
```
tsa96, line 36: Does this need to be `useClass` rather than `useValue`? Isn't SentryInterceptor a constant value? 

https://docs.nestjs.com/fundamentals/custom-providers#class-providers-useclass

server/src/app.module.ts
```diff
@@ -1,29 +1,58 @@
...
+            dsn: appConfig.sentry.dsn,
+            tracesSampleRate: 1.0,
+            debug: appConfig.sentry.debug,
+            environment: process.env.NODE_ENV || 'development'
+        })
+    ],
+    providers: [
+        {
+            provide: APP_FILTER,
+            useClass: AllExceptionsFilter
```
tsa96, line 32: Same as above

tsa96 (changes requested):
One thing, and I might be misunderstanding the custom provider stuff, feel free to mark resolved if so

server/src/@common/sentry/sentry.module.ts
TheNerdMan: UseValue is used when you want to provide a static Object, use class is used when you want to provide a static Service or similar that has a constructor that needs to be run for the service to be useful.

https://medium.com/@paynoattn/use-usevalue-if-your-stub-is-a-object-useclass-if-your-stub-is-a-class-88ba6e1574f0
https://offering.solutions/blog/articles/2018/08/17/using-useclass-usefactory-usevalue-useexisting-with-treeshakeable-providers-in-angular/

server/src/app.module.ts
TheNerdMan: UseValue is used when you want to provide a static Object, use class is used when you want to provide a static Service or similar that has a constructor that needs to be run for the service to be useful.

https://medium.com/@paynoattn/use-usevalue-if-your-stub-is-a-object-useclass-if-your-stub-is-a-class-88ba6e1574f0
https://offering.solutions/blog/articles/2018/08/17/using-useclass-usefactory-usevalue-useexisting-with-treeshakeable-providers-in-angular/

tsa96 (approved):

Gocnak (approved):

PR #634, opened 2022-06-05
Safer app bootstrapping

TheNerdMan:
This allows our bootstrapping of the app to be as similar as possible
between production and jest testing. 

It should lower risk of errors showing in on version of a build (eg in test but not production) by properly enabling settings in predefined functions.


tsa96 (approved):
Okay, I've tested the BigInt stuff properly and it's working

PR #637, opened 2022-06-15
User/ endpoint work & Stub remaining endpoints

tsa96:
Completes most stuff in user/.

Then adds controller and DTO layer for all the remaining endpoints I could find besides maps/ (Alex) and some remaining on auth/ (me, next).

This should make life a lot easier for our new devs having an initial layer in place, requires referring back to old API much less.


server/src/@common/dto/map/map-notify.dto.ts
```diff
@@ -0,0 +1,22 @@
+﻿import { ApiProperty, PickType } from '@nestjs/swagger';
+import { MapNotify } from '@prisma/client';
+import { Exclude } from 'class-transformer';
+
+export class MapNotifyDto implements MapNotify {
```
TheNerdMan, line 5: Missing properties that are on the Prisma model

server/src/@common/dto/map/createMap.dto.ts
```diff
@@ -1,7 +1,8 @@
 import { ApiProperty } from '@nestjs/swagger';
 import { EMapTriggerType, EMapType } from '../../enums/map.enum';
-import { MapCreditDto } from './mapCredit.dto';
+import { MapCreditDto } from './map-credit.dto';
 
+// TODO: Alex, these are really messy, can they be moved to their respective DTO files?
```
TheNerdMan, line 5: Yep yep, this was just temp while I got the tests working

server/src/@common/filters/all-exceptions.filter.ts
```diff
@@ -30,7 +31,7 @@ export class AllExceptionsFilter implements ExceptionFilter {
             e
         );
 
-        if (!(e instanceof UnauthorizedException) && !(e instanceof NotFoundException)) {
+        if (e instanceof InternalServerErrorException) {
```
TheNerdMan, line 34: Has this been tested that it catches all errors appart from 401 and 404?

server/src/@common/dto/map/map-notify.dto.ts
tsa96: I think your Prisma client is probably out of date, the current type is
```ts
export type MapNotify = {
  notifyOn: number
  mapID: number
  userID: number
  createdAt: Date
  updatedAt: Date
}
```

server/src/@common/filters/all-exceptions.filter.ts
tsa96: We don't want it catching any of the 400s. Some other 500s we should probably get, I'll come back to this

Gocnak (approved):

PR #645, opened 2022-07-08
runs/ and runs/:mapID endpoints

benjl:


server/src/modules/runs/runs.service.ts
```diff
@@ -1,7 +1,45 @@
...
+        if (mapID) {
+            where.mapID = mapID;
+        }
+        if (userID) {
+            where.playerID = userID;
+        } else if (userIDs) {
+            where.playerID = { in: userIDs };
+        }
+        if (flags) {
+            where.flags = flags; // TODO: Change this to bitwise enum crap
```
tsa96, line 54: Let's leave a comment here mentioning that this will be expanded during 0.10.0. But not as a TODO, those should be for things within the rewrite only.

server/tests/runs.e2e-spec.ts
```diff
@@ -0,0 +1,341 @@
...
+                file: 'world_record.run',
+                hash: '0xdeadbeef',
+                rank: {
+                    create: {
+                        gameType: MapType.RJ,
+                        flags: 0,
+                        trackNum: 0,
+                        zoneNum: 0,
+                        rank: 6,
+                        createdAt: new Date(),
```
tsa96, line 120: Why the `createdAt `here? This is handled by Prisma (note `creationDate` on `Map`s is an actual date we track ourselves, since that could be for some map originally released years ago) . Same with other `Run` creations in here.

server/tests/runs.e2e-spec.ts
```diff
@@ -0,0 +1,341 @@
...
+                },
+                baseStats: {
+                    create: {
+                        jumps: 1
+                    }
+                },
+                zoneStats: {
+                    createMany: {
+                        data: [
+                            {zoneNum: 1},
```
tsa96, line 133: Formatting looks off here, are you sure you've run Prettier?

server/tests/runs.e2e-spec.ts
```diff
@@ -0,0 +1,341 @@
...
+        await prisma.map.deleteMany({ where: { id: { in: [map1.id, map2.id] } } });
+    });
+
+    describe('GET /api/v1/runs', () => {
+        const expects = (res) => expect(res.body).toBeValidPagedDto(RunDto);
+        const filter = (x) => {[run1.id, run2.id, run3.id, run4.id].includes(x.id)};
+        it('should respond with a list of runs', async () => {
+            const res = await get('runs', 200);
+            expects(res);
+            expect(res.body.totalCount).toBeGreaterThan(1);
```
tsa96, line 280: Lower bound here is 4, no?

server/tests/runs.e2e-spec.ts
```diff
@@ -0,0 +1,341 @@
...
+            expect(res.body.totalCount).toBe(4);
+            expect(res.body.returnCount).toBe(4);
+            expect(ids).toContain(res.body.response[0].playerID.toString());
+        });
+
+        it('should respond with a list of runs filtered by flags', async () => {
+            const res = await get('runs', 200, { flags: 1 << 5 });
+            expects(res);
+            expect(res.body.totalCount).toBe(1);
+            expect(res.body.response[0].flags).toBe(run4.flags); // TODO: Change this to bitwise math later
```
tsa96, line 317: Similar to above comment, lets remove the TODO here since it's not really a rewrite thing, replace with a note it'll change in 0.10.0. Just doing a strict equality check like this is fine for now.

server/tests/runs.e2e-spec.ts
```diff
@@ -0,0 +1,341 @@
...
+        it('should respond with a list of runs with the zoneStats include',  () => expandTest('runs', expects, 'zoneStats', true));
+
+        it('should respond with a list of runs with the baseStats include', () => expandTest('runs', expects, 'baseStats', true));
+
+        it('should respond with a list of runs with the mapWithInfo include', async () => {
+            const res = await get('runs', 200, { expand: 'mapWithInfo' });
+            expects(res);
+            res.body.response.forEach(x => {
+                expect(x.map.info).not.toBeNull();
+            });
```
tsa96, line 335: I prefer we use `expect(x.map).toHaveProperty('info')` here, mainly just to keep consistency with other tests. Also the curly braces are unnecessary!

server/tests/runs.e2e-spec.ts
```diff
@@ -0,0 +1,341 @@
...
+        it('should respond with a list of runs filtered by userID parameter', async () => {
+            const res = await get('runs', 200, { userID: user1.id });
+            expects(res);
+            expect(res.body.totalCount).toBe(2);
+            expect(res.body.returnCount).toBe(2);
+            expect(res.body.response[0].playerID).toBe(user1.id);
+        });
+
+        it('should respond with a list of runs filtered by a list of run ids', async () => {
+            const ids: String = user1.id + ',' + user2.id;
```
tsa96, line 305: Don't need the `String` type here, can be easily inferred from the RHS (also it should be `string`).

server/tests/runs.e2e-spec.ts
```diff
@@ -338,4 +338,34 @@ describe('runs', () => {
...
+        it('should respond with a run using the map include', () => expandTest('runs/' + run1.id, expects, 'map', false));
+
+        it('should respond with a run using the rank include', () => expandTest('runs/' + run1.id, expects, 'rank', false));
+
+        it('should respond with a run using the zoneStats include', () => expandTest('runs/' + run1.id, expects, 'zoneStats', false));
+
+        it('should respond with a run using the mapWithInfo include', async () => {
+            const res = await get('runs/' + run1.id, 200, { expand: 'mapWithInfo' });
+            expects(res);
+            expect(res.body.map.info).not.toBeNull();
```
tsa96, line 364: Let's use `toHaveProperty`

server/tests/runs.e2e-spec.ts
```diff
@@ -338,4 +338,34 @@ describe('runs', () => {
         it('should respond with 401 if no access token is provided', () => get('runs', 401, {}, null));
 
     });
+
+    describe('GET /api/v1/runs/{runID}', () => {
+        const expects = (res: request.Response) => expect(res.body).toBeValidDto(RunDto);
+        it('should return a valid run', async () => {
+            const res = await get('runs/' + run1.id, 200);
+            expects(res);
+            expect(res.body.id).toBe(run1.id.toString()); // The id is a string for some reason, change when that's fixed
```
tsa96, line 347: This is a JSON response, so everything is a string! The `toString` call here is fine, you can just remove the comment.

server/src/modules/runs/runs.service.ts
```diff
@@ -1,14 +1,29 @@
...
+        const where: Prisma.RunWhereUniqueInput = { id: runID };
+        const include: Prisma.RunInclude = ExpandToPrismaIncludes(
+            expand?.filter((x) => ['baseStats', 'map', 'player', 'rank', 'zoneStats'].includes(x))
+        );
+
+        if (expand?.includes('mapWithInfo')) include.map = { include: { info: true } };
+
+        const dbResponse = await this.runRepo.get(where, include);
+
+        if (!dbResponse) throw new NotFoundException();
```
tsa96, line 22: Include a message i.e. "Run not found"

tsa96 (changes requested):
Great start! Most stuff is okay, just various tiny little things. And don't be put off by the more formal review-talk, personally I find it weird going between this and casual chat on Discord, we just like to keep our reviews more formal.

There's a few things on the old API I see that look like they're missing (this is to be expected, the runs/get endpoint has a ton of stuff!).
getAll
- Uses the distinct option to filter out any duplicates (see https://www.prisma.io/docs/concepts/components/prisma-client/aggregation-grouping-summarizing#select-distinct)
- Has a default `take` of 10, we use 20 by default  in `PaginationQuery` so you wanna override that
- Ordering. Not gonna lie, this is confusing me quite a lot, lets discuss on Discord.
- mapName filter. Seems to use a "like" comparison, idk what that is, again we can talk about it.
- There's some really really gross stuff on L706 for `isPB`

Like I said the other day, apologies for the gross stuff on the old API, it can be a real pain to read through. I'm happy to try figure stuff out for you if you can't make sense of it (at worst we just ask Goc).

Lastly, make sure you rebase fixes here then force push, we never want to muddy the main branch up with review revisions. As always, happy to help with that in Discord.

tsa96:
Oh yeah, and getAll is always including the User object

server/tests/runs.e2e-spec.ts
tsa96: Never mind I'm an idiot. It's cus it's a bigint we serialise out to string

server/tests/runs.e2e-spec.ts
```diff
@@ -0,0 +1,376 @@
...
+            expect(res.body.returnCount).toBe(2);
+            res.body.response.forEach(x => {
+                expect(x).toHaveProperty('rank');
+            });
+        })
+
+        it('should respond with a list of runs sorted by date', async () => {
+            const res = await get('runs', 200, { order: 'date'});
+            expects(res);
+            const sortedRes = [...res.body.response];
```
tsa96, line 327: Just a thing to keep in mind here (though it's fine in this case) - this method of object cloning is a *shallow clone*; the response objects themselves will be the same object here, the order is just different. So what you're doing is completely fine, just an important thing to understand as you learn more JS. https://developer.mozilla.org/en-US/docs/Glossary/Deep_copy for more info

server/tests/runs.e2e-spec.ts
```diff
@@ -0,0 +1,376 @@
...
+        const prisma: PrismaService = global.prisma;
+
+        await prisma.user.deleteMany({ where: { id: { in: [user1.id, user2.id] } } });
+        await prisma.run.deleteMany({ where: { id: { in: [run1.id, run2.id, run3.id, run4.id] } } });
+        await prisma.map.deleteMany({ where: { id: { in: [map1.id, map2.id] } } });
+    });
+
+    describe('GET /api/v1/runs', () => {
+        const expects = (res) => expect(res.body).toBeValidPagedDto(RunDto);
+        const filter = (x) => {[run1.id, run2.id, run3.id, run4.id].includes(x.id)};
```
tsa96, line 241: What's the filter being used for? Looks like it should be being used in the `expandTest`s

server/src/modules/runs/runs.service.ts
```diff
@@ -1,7 +1,72 @@
...
+        else if (mapName) where.map = { name: { contains: mapName } };
+
+        if (userID) where.playerID = userID;
+        else if (userIDs) where.playerID = { in: userIDs };
+
+        if (isPB) where.rank = { isNot: null };
+
+        if (flags) where.flags = flags; // Currently checks for exact equality, will change in 0.10.0
+
+        if (order == 'date') orderBy.createdAt = 'desc';
```
tsa96, line 66: Strict compare `===`

server/src/@common/dto/run/runs.dto.ts
```diff
@@ -63,6 +65,16 @@ export class RunDto implements Run {
     @ApiProperty()
     hash: string;
 
+    @ApiProperty({ type: () => BaseStatsDto })
```
tsa96, line 68: No need for lazy reference

tsa96 (changes requested):

tsa96 (approved):

tsa96:
Congratulations you are now a web developer

PR #648, opened 2022-07-28
More maps endpoints

benjl:
We got the following endpoints:
- maps/:mapID/credits GET, POST  
- maps/credits/:mapCredID GET, PATCH, DEL  
- maps/:mapID/info GET, PATCH  
- maps/:mapID/zones GET  

I decided not to bother messing around moving any service logic, imo it's probably best to do all of that in one go anyways

### DIFF
--- a/server/src/@common/dto/map/map-info.dto.ts
+++ b/server/src/@common/dto/map/map-info.dto.ts
@@ -1,6 +1,6 @@
 ﻿import { MapInfo } from '@prisma/client';
 import { Exclude, Transform, Type } from 'class-transformer';
-import { ApiProperty, PickType } from '@nestjs/swagger';
+import { ApiProperty, PartialType, PickType } from '@nestjs/swagger';
 import { IsDate, IsDateString, IsDefined, IsInt, IsOptional, IsString, Matches } from 'class-validator';
 
 export class MapInfoDto implements MapInfo {
@@ -45,3 +45,5 @@ export class CreateMapInfoDto extends PickType(MapInfoDto, [
     'numTracks',
     'creationDate'
 ] as const) {}
+
+export class UpdateMapInfoDto extends PartialType(PickType(MapInfoDto, ['description', 'youtubeID', 'creationDate'] as const)) {}
--- a/server/src/modules/maps/maps.service.ts
+++ b/server/src/modules/maps/maps.service.ts
@@ -16,6 +16,7 @@ import { DtoFactory, ExpandToPrismaIncludes } from '../../@common/utils/dto.util
 import { UsersRepoService } from '../repo/users-repo.service';
 import { ActivityTypes } from '../../@common/enums/activity.enum';
 import { CreateMapCreditDto, MapCreditDto, UpdateMapCreditDto } from '../../@common/dto/map/map-credit.dto';
+import { MapInfoDto, UpdateMapInfoDto } from '../../@common/dto/map/map-info.dto';
 
 @Injectable()
 export class MapsService {
@@ -375,6 +376,36 @@ export class MapsService {
         await this.mapRepo.deleteCredit({ id: mapCredID });
 
         await this.updateActivities(null, mapCred);
+    }
+    //#endregion
+
+    //#region Map Info
+    async getInfo(mapID: number): Promise<MapInfoDto> {
+        const mapInfo = await this.mapRepo.getInfo(mapID);
+
+        if (!mapInfo) throw new NotFoundException('Map not found');
+
+        return DtoFactory(MapInfoDto, mapInfo);
+    }
+
+    async updateInfo(mapID: number, mapInfo: UpdateMapInfoDto, userID: number): Promise<void> {
+        if (!mapInfo.description && !mapInfo.youtubeID && !mapInfo.creationDate)
+            throw new BadRequestException('Request contains no valid update data');
+        const map = await this.mapRepo.get(mapID);
+        if (!map) throw new NotFoundException('Map not found');
+
+        if (map.submitterID !== userID) throw new ForbiddenException('User is not the submitter of the map');
+
+        if (map.statusFlag !== MapStatus.NEEDS_REVISION)
+            throw new ForbiddenException('Map is not in NEEDS_REVISION state');
+
+        const data: Prisma.MapInfoUpdateInput = {};
+
+        if (mapInfo.description) data.description = mapInfo.description;
+        if (mapInfo.youtubeID) data.youtubeID = mapInfo.youtubeID;
+        if (mapInfo.creationDate) data.creationDate = new Date(mapInfo.creationDate);
+
+        await this.mapRepo.updateInfo(mapID, data);
     }
     //#endregion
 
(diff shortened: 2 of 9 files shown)


server/src/modules/maps/maps.service.ts
```diff
@@ -286,6 +287,56 @@ export class MapsService {
...
+        if (!foundMap) throw new NotFoundException('Map not found');
+
+        const include: Prisma.MapCreditInclude = ExpandToPrismaIncludes(expand?.filter((x) => ['user'].includes(x)));
+        const where: Prisma.MapCreditWhereInput = { mapID: mapID };
+
+        const dbResponse = await this.mapRepo.getCredits(where, include);
+
+        if (dbResponse.length === 0) throw new NotFoundException('No map credits found');
+
+        const mapCredits = dbResponse.map((x) => DtoFactory(MapCreditDto, x));
```
tsa96, line 304: No need for assignment, just return

server/src/@common/dto/map/map-info.dto.ts
```diff
@@ -45,3 +45,20 @@ export class CreateMapInfoDto extends PickType(MapInfoDto, [
...
+    @ApiProperty({ description: 'The new youtube ID to set', type: String })
+    @IsOptional()
+    @Matches(/^[\w_-]{11}$/)
+    youtubeID: string;
+
+    @ApiProperty( { description: 'The new creation date to set', type: Date })
+    @IsOptional()
+    @IsDateString()
+    creationDate: Date;
+}
```
tsa96, line 64: I think it's best we extend the existing types on the MapInfoDto - I don't even mind that it won't have a description specifically about updating. So we can use [NestJS's type mapping](https://docs.nestjs.com/openapi/mapped-types) and simply do something like
```ts
export class UpdateMapInfoDto extends PartialType(PickType(MapInfoDto, ['description', 'youtubeID', 'creationDate'] as const)) {}
```

server/src/modules/maps/maps.service.ts
```diff
@@ -380,6 +381,35 @@ export class MapsService {
...
+        if (!mapInfo) throw new NotFoundException('Map not found');
+
+        return DtoFactory(MapInfoDto, mapInfo);
+    }
+
+    async updateInfo(mapID: number, mapInfo: UpdateMapInfoDto, userID: number): Promise<void> {
+        if (!mapInfo.description && !mapInfo.youtubeID && !mapInfo.creationDate)
+            throw new BadRequestException('Request contains no valid update data');
+        const map = await this.mapRepo.get(mapID);
+        if (!map) throw new NotFoundException('Map not found');
```
tsa96, line 397: nitpicky but feels like there should be spaces between these lines

server/src/modules/repo/maps-repo.service.ts
```diff
@@ -142,6 +142,23 @@ export class MapsRepoService {
...
     //#endregion
 
+    //#region MapInfo
+    async getInfo(mapId: number): Promise<MapInfo> {
+        const mapInfo = await this.prisma.mapInfo.findUnique({ where: { mapID: mapId } });
+
+        return mapInfo;
+    }
+
+    async updateInfo(mapId: number, data: Prisma.MapInfoUpdateInput): Promise<MapInfo> {
```
tsa96, line 152: `mapID` not `mapId`

server/src/modules/maps/maps.service.ts
```diff
@@ -410,6 +411,24 @@ export class MapsService {
...
+
+        const tracks = await this.mapRepo.getMapTracks(mapID);
+
+        // This is dumb but it's what the old api does
+        // \server\src\models\map.js Line 499
+        // TODO: When map sessions are done this should be removed
+
+        await this.mapRepo.updateMapStats(mapID, {
+            plays: ((map as any).stats as MapStats).plays + 1 ?? 1
```
tsa96, line 426: Should use [increment](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#atomic-number-operations) here i.e.
```ts
plays: { increment: 1 }
```
 Prisma schema defaults `plays` to 0 so we don't have to worry about it being undefined.

server/src/modules/repo/maps-repo.service.ts
```diff
@@ -159,7 +159,36 @@ export class MapsRepoService {
...
+                        triggers: {
+                            include: {
+                                properties: true
+                            }
+                        }
+                    }
+                }
+            }
+        });
+    }
```
tsa96, line 191: We should be able to pass in a Prisma include here, we don't want this to hardcode the include. In some cases like creates I don't mind always returning the same include, but this seems like something likely to be used in a bunch of places and likely to want various different includes.

server/src/modules/maps/maps.service.ts
```diff
@@ -410,6 +411,24 @@ export class MapsService {
...
+    async getZones(mapID: number): Promise<MapTrackDto[]> {
+        const map = await this.mapRepo.get(mapID, { stats: true });
+
+        if (!map) throw new NotFoundException('Map not found');
+
+        const tracks = await this.mapRepo.getMapTracks(mapID);
+
+        // This is dumb but it's what the old api does
+        // \server\src\models\map.js Line 499
+        // TODO: When map sessions are done this should be removed
```
tsa96, line 423: Please use `TODO_POST_REWRITE`, `TODO` should only be for rewrite stuff

tsa96 (changes requested):
Amazing work. Relatively few issues for a PR of this size, I'm really impressed. Very happy you've been able to handle some of the more complicated service logic, and your tests are great, way better than stuff on the old API.

server/src/modules/maps/maps.service.ts
```diff
@@ -286,6 +287,56 @@ export class MapsService {
...
+        const mapCredits = dbResponse.map((x) => DtoFactory(MapCreditDto, x));
+
+        return mapCredits;
+    }
+
+    async createCredit(mapID: number, createMapCredit: CreateMapCreditDto, userID: number): Promise<MapCreditDto> {
+        const map = await this.mapRepo.get(mapID, { credits: true });
+        if (!map) throw new NotFoundException('Map not found');
+
+        if (map.submitterID != userID) throw new ForbiddenException('User is not the submitter of the map');
```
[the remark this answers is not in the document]
tsa96: There's a couple other cases of non-strict checks as well

server/tests/maps.e2e-spec.ts
```diff
@@ -929,41 +973,104 @@ describe('Maps', () => {
...
-            const res = await get('maps/999999999999999/credits', 200);
-
-            expect(res.body.mapCredits).toHaveLength(0);
+        it('should respond with the specified maps credits with the user expand parameter', async () => {
+            const res = await get(`maps/${map1.id}/credits`, 200, { expand: 'user' });
+            expects(res);
+            res.body.forEach((x) => {
+                expect(x).toHaveProperty('user');
+                expect(x.user).toBeValidDto(UserDto);
+            });
```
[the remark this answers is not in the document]
benjl: `expects` won't complain if the user property isn't defined, no? Since `user` is an expand parameter, it shouldn't be defined in the test above this one which would mean `expects` would fail there, so the `forEach` here should specifically check that `user` is there.

tsa96 (approved):

PR #650, opened 2022-08-12
MapDto and map upload test bugfixes

benjl:


tsa96 (approved):
Nice, thanks for fixing these!

Issue #620, opened 2022-03-29
Hook up Sentry for error reporting


Gocnak:
NestJS will specifically need https://www.npmjs.com/package/@ntegral/nestjs-sentry
Angular has its own official guide: https://docs.sentry.io/platforms/javascript/guides/angular/

We're on Sentry for error reporting for the game and should absolutely use it for the front and backend. There's even [performance checking, too.](https://docs.sentry.io/product/performance/)

We should also update our main readme to say we utilize it, and plop their logo in: https://sentry.io/branding/

tsa96:
Module is in place, but not tested live - check this out after prod

tsa96:
`exception-hander.filter.ts` has some weird looking logic - it appears that *all* errors are going to get set to Sentry when in prod. Fix that when working on this!

tsa96:
Almost done, just need to get sourcemaps working.

tsa96:
Pretty much, syuff we can improve but that'll be an ongoing thing.

PR #651, opened 2022-08-24
#620 feat(sentry): sentry integration for a client

sergeevNick:
Need to add sentry DSN to envs.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -51,6 +51,8 @@
     "@ngx-gallery/gallerize": "4.0.3",
     "@ngx-gallery/lightbox": "4.0.3",
     "@node-steam/vdf": "2.1.0",
+    "@sentry/angular": "7.11.1",
+    "@sentry/tracing": "7.11.1",
     "@trademe/ng-defer-load": "^8.2.1",
     "bootstrap": "4.5.0",
     "classlist.js": "1.1.20150312",
--- a/client/src/environments/environment.prod.ts
+++ b/client/src/environments/environment.prod.ts
@@ -7,4 +7,5 @@ export const environment = {
   production: true,
   api: 'https://api.momentum-mod.org',
   auth: 'https://auth.momentum-mod.org',
+  sentry: { dsn: '', env: 'client-production' },
 };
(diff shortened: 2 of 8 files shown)


client/package.json
```diff
@@ -51,8 +51,8 @@
     "@ngx-gallery/gallerize": "4.0.3",
     "@ngx-gallery/lightbox": "4.0.3",
     "@node-steam/vdf": "2.1.0",
-    "@sentry/angular": "^7.11.1",
-    "@sentry/tracing": "^7.11.1",
+    "@sentry/angular": "7.11.1",
+    "@sentry/tracing": "7.11.1",
```
tsa96, line 55: This is in a separate commit for some reason, please fixup into the other and force-push (we always make sure to keep the master branch all of Momentum repo's as clean as possible, even if that requires other branches are regularly force-pushed to)

client/src/environments/environment.prod.ts
```diff
@@ -7,4 +7,5 @@ export const environment = {
   production: true,
   api: 'https://api.momentum-mod.org',
   auth: 'https://auth.momentum-mod.org',
+  sentry: { dsn: '', env: 'client-production' },
```
tsa96, line 10: Even assuming we're happy with the fact we pass the Sentry DSN to the client, I don't think that should mean we store it publicly on Github where it's more visible and potentially scraped and spammed - I dunno what/who would, but seems like an unnecessary risk. Better to have it loaded from the .env file, it's already defined in there on the NestJS branch as `SENTRY_DSN` in `.env`.

tsa96 (changes requested):
Looks like you've done all the actual Sentry stuff really well, just got to clean up the double commit and load the DSN from .env - we can discuss either in Discord anytime if you want.

I'm not going to merge this myself, I'd rather wait for Goc to have a look, which may not be until rewrite is done and we get Sentry on the backend up and running, though that's up to him. So rest assured if this takes a bit to get merged we've not forgotten about it and it'll get in round the time rewrite is done - in fact it's great to have this out of the way. Thanks for the first PR!!

client/package.json
sergeevNick: Done

tsa96:
Not planning on doing Sentry for frontend

PR #663, opened 2022-11-09
nestJS activities endpoint

TheJazzAgenT:
nestJS activities endpoint
`api/v1/activities`

Added e2e activities tests in.


server/tests/e2e/activities.e2e-spec.ts
```diff
@@ -0,0 +1,200 @@
...
+                steamID: '754673452345',
+                alias: 'User 3',
+                roles: { create: { mapper: true } }
+            }
+        });
+
+
+        map1 = await prisma.map.create( {
+            data: {
+                name: 'maps_test' + Math.floor(Math.random() * 100000000),
```
tsa96, line 38: Lets change naming here, not just copy verbatim from maps test. Also looks like this could just be a hardcoded name?

tsa96: Actually looks like map1 is unneeded

server/tests/e2e/activities.e2e-spec.ts
```diff
@@ -0,0 +1,200 @@
...
+
+    describe('GET /api/v1/activities', () => {
+        const expects = (res: request.Response) => expect(res.body).toBeValidPagedDto(ActivityDto);
+        it('should respond with an array of activities', async () => {
+            const res = await get('activities', 200);
+
+            expects(res);
+            expect(res.body.totalCount).toBe(3);
+            expect(res.body.response[0].userID).toBe(user1.id);
+            expect(res.body.response[1].userID).toBe(user1.id);
```
tsa96, line 147: We can improve this test a little by also checking the `type` `data` values are right.

That being said, this test looks like it's going to depend on the order of activities being inserted. Also, if we're running Jest concurrently and there's some other test inserting activities, this data could be wrong.

I think a reasonable approach would be to first `.filter` the response data by say the user ID, then check the data and types are correct.  So how about

```ts
expect(res.body.totalCount).toBeGreaterThanOrEqual(3);
expect(res.body.response.find((data) => data.userID === user1.id && data.data === 122).type).toBe(ActivityTypes.MAP_APPROVED);
expect(res.body.response.find((data) => data.userID === user1.id && data.data === 123).type).toBe(ActivityTypes.WR_ACHIEVED);
expect(res.body.response.find((data) => data.userID === user2.id).type).toBe(ActivityTypes.REVIEW_MADE);
```

server/tests/e2e/activities.e2e-spec.ts
```diff
@@ -0,0 +1,200 @@
...
+            expect(res.body.totalCount).toBe(1);
+            expect(res.body.response[0].userID).toBe(user2.id);
+            expect(res.body.response[0].data).toBe('124');
+        });
+
+        it('should respond with array of activities with type paramater', async () => {
+            const res = await get('activities', 200, {type: 'MAP_APPROVED'});
+
+            expect(res);
+            expect(res.body.totalCount).toBe(1);
```
tsa96, line 173: Could be different when running concurrently, lets use toBeGreaterThanOrEqualTo or whatever it's called

tsa96: But maybe best to make sure the test is still doing a strong enough check, so let's also do
```ts
expect(res.body.response.filter((data) => data.userID === user1.id || data.userID === user2.id).length).toBe(1).
```
Quite likely we switch to --runInBand for everything in the future but lets keep this up for now.

server/tests/e2e/activities.e2e-spec.ts
```diff
@@ -0,0 +1,200 @@
...
+        });
+
+        it('should respond with array of all activities with type ALL paramater', async () => {
+            const res = await get('activities', 200, {type: 'ALL'});
+
+            expect(res);
+            expect(res.body.totalCount).toBe(3);
```
tsa96, line 180: Same as above comment

tsa96 (changes requested):
This is looking great, really impressed with how fast you got into this. Only additional thing is to run Prettier on everything please

server/src/modules/activities/activities.service.ts
```diff
@@ -1,7 +1,32 @@
...
     constructor(private readonly userRepo: UsersRepoService) {}
+
+    async getAll(query: ActivitiesGetQuery): Promise<PaginatedResponseDto<ActivityDto>> {
+        const where: Prisma.ActivityWhereInput = {};
+        if (query.userID) {
+            where.userID = query.userID;
+        }
+
+        // if type is ALL, just don't add a type filter
+        if (query.type && Number(ActivityTypes[query.type]) !== 0) {
```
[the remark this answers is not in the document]
tsa96: This still needs doing

tsa96 (changes requested):

PR #665, opened 2022-11-16
feat: nestJS/user/maps/library/mapid/GET

TheJazzAgenT:
nestJS GET /user/maps/library/{mapid} endpoint

### DIFF
--- a/server/src/modules/maps/map-library.service.ts
+++ b/server/src/modules/maps/map-library.service.ts
@@ -1,0 +1,40 @@
+import {
+    BadRequestException,
+    Injectable,
+    NotFoundException
+} from '@nestjs/common';
+import { PrismaService } from '../repo/prisma.service';
+import {
+    Prisma
+} from '@prisma/client';
+
+
+@Injectable()
+export class MapLibraryService {
+    constructor( private prisma: PrismaService ) {}
+
+    async isMapInLibrary(userID: number, mapID: number) {
+        const mapWhere: Prisma.MapWhereInput = {
+            id: mapID
+        };
+
+        const map = await this.prisma.map.findFirst({
+            where: mapWhere
+        });
+
+        if ( !map ) {
+            throw new BadRequestException('Map does not exist');
+        }
+
+        const where: Prisma.MapLibraryEntryWhereInput = {
+            userID: userID,
+            mapID: mapID
+        };
+
+        const dbResponse = await this.prisma.mapLibraryEntry.findFirst({
+            where: where,
+        });
+
+        if (!dbResponse) throw new NotFoundException('Map not found');
+    }
+}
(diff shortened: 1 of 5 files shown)


server/tests/e2e/user.e2e-spec.ts
```diff
@@ -754,34 +754,21 @@ describe('User', () => {
...
-    //                 .expect(401)
-    //                 .expect('Content-Type', /json/));
-    //     });
+    
+    describe('GET /api/v1/user/maps/library/{mapID}', () => {
+        it('should check if a map exists in the local users library', () => {
+            getNoContent(`user/maps/library/${map1.id}`, 204);
+        });
+
+        it('should return 404 since the map is not in the local users library', () => {
```
tsa96, line 763: Nit, but we describe these with *if* the map, not since.

server/src/modules/user/user.controller.ts
```diff
@@ -48,7 +49,7 @@ import { MapFavoriteDto } from '../../common/dto/map/map-favorite.dto';
 @Controller('api/v1/user')
 @ApiTags('User')
 export class UserController {
-    constructor(private readonly usersService: UsersService) {}
+    constructor(private readonly usersService: UsersService, private readonly MapLibraryService: MapLibraryService) {}
```
tsa96, line 52: Lower camel case please i.e. `private readonly mapLibraryService: MapLibraryService`

server/src/modules/maps/map-library.service.ts
```diff
@@ -0,0 +1,40 @@
...
+
+
+@Injectable()
+export class MapLibraryService {
+    constructor( private prisma: PrismaService ) {}
+
+    async isMapInLibrary(userID: number, mapID: number) {
+        const mapWhere: Prisma.MapWhereInput = {
+            id: mapID
+        };
```
tsa96, line 19: Declaring this separately feels redundant, just for the findFirst just do `where: { id: mapID }`

server/src/modules/maps/map-library.service.ts
```diff
@@ -0,0 +1,40 @@
...
+
+        const map = await this.prisma.map.findFirst({
+            where: mapWhere
+        });
+
+        if ( !map ) {
+            throw new BadRequestException('Map does not exist');
+        }
```
tsa96, line 27: No need for curly braces around a one-liner that's just a `throw` (I generally avoid braces around one-liners everywhere). Also, since you've done `( !map )` instead of `(!map)` you've definitely not run Prettier. I'll have better guidelines on that very soon, for now can you just install the VSCode extension and set that as your formatter, then run the Format File command or whatever it's called

tsa96 (changes requested):
Looking good, few nits. In the future, now you're starting to get up to speed with everything, I'd say it's generally not PRing a single endpoint unless it's really complex. Assuming you're gonna work on further endpoints at least - lets definitely do the rest of the map library stuff in one PR, if you're up for working on that.

PR #679, opened 2023-01-01
feat: PUT /api/user/maps/favorites/{mapID}

BossCode45:


server/src/modules/repo/users-repo.service.ts
```diff
@@ -297,6 +297,15 @@ export class UsersRepoService {
         return [mapLibraryEntries, count];
     }
 
+	async addMapLibraryEntry(userID: number, mapID: number) {
```
tsa96: Lets get naming consistent with rest of repo service and use `createMapLibraryEntry`.

server/src/modules/user/user.controller.ts
```diff
@@ -354,9 +354,8 @@ export class UserController {
     })
     @ApiNoContentResponse({ description: 'Map was added to the library' })
     @ApiNotFoundResponse({ description: 'The map does not exist' })
-    addMapLibraryEntry(@LoggedInUser('id') _userID: number, @Param('mapID', ParseIntPipe) _mapID: number) {
-        return void 0;
-        //return this.usersService.AddMapLibraryEntry(userID, mapID);
+    addMapLibraryEntry(@LoggedInUser('id') _userID: number, @Param('mapID', ParseIntPipe) _mapID: number){
+        return this.usersService.addMapLibraryEntry(_userID, _mapID);
```
tsa96, line 358: I was underscoring the variable names here as they were unused (so linter would shout at us), but now they're being used, no need for the underscores

server/tests/e2e/user.e2e-spec.ts
```diff
@@ -1115,6 +1115,45 @@ describe('User', () => {
         });
     });
 
+	describe('PUT /api/user/maps/library/{mapID}', () => {
+		it('should add a new map to the local users library', async () => {
+			const mapLibraryEntry = await global.prisma.mapLibraryEntry.findFirst({
```
tsa96: Some editors will complain here that the type of `global.prisma` is unknown so not know what `mapLibraryEntry` is. So please specific explicitly with `(global.prisma as PrismaService).mapLibraryEntry...`

server/tests/e2e/user.e2e-spec.ts
```diff
@@ -1115,6 +1115,45 @@ describe('User', () => {
...
+				where: {
+					userID: user1.id,
+					mapID: map3.id }
+				}
+			);
+			expect(mapLibraryEntry).toBeNull();
+
+			await put({
+				url: `user/maps/library/${map3.id}`,
+				status: 200,
```
tsa96: This endpoint should be 204ing. Looks like missing a `@HttpCode(HttpStatus.NO_CONTENT)` below the `@Put` declaration, my bad.

server/src/modules/user/user.controller.ts
```diff
@@ -45,7 +45,7 @@ import { MapDto } from '@common/dto/map/map.dto';
 import { MapFavoriteDto } from '@common/dto/map/map-favorite.dto';
 import { MapLibraryService } from '@modules/maps/map-library.service';
 
-@Controller('api/user')
+@Controller('user')
```
tsa96, line 48: This isn't the right fix for the tests 404ing - I'll have that in shortly. So please leave this out of the commit, if you want to fix everything in the meantime, just remove L42 `app.setGlobalPrefix('api', { exclude: ['auth'] });` from `tests/e2e/environment.ts`.

tsa96 (changes requested):
Logic and tests looks good, nice work, just got various smaller issues.

One important thing with Momentum's approach to Git: we never want commits making fixes to your current (non-master) branch merged into master, instead just do a interactive rebase and `fixup`. So the styling and linting commit shouldn't exist at all.

I must admit, last night I *did* accidentally merge in my branch before styling checks are complete so needed to do a commit specifically for styling, but generally that's something we really want to avoid. Also, since your styling commit made some changes to `auth.controller.ts` and shit I fixed in my styling fix, it looks like you're not on top of that branch - just fetching and rebasing your branch on top of that should resolve everything fine.

tsa96 (approved):

PR #682, opened 2023-01-01
feat: DELETE /api/user/maps/library/{mapID} DELETE & PUT /api/user/ma…

BossCode45:
…ps/favorites/{mapID}

### DIFF
--- a/server/src/modules/repo/users-repo.service.ts
+++ b/server/src/modules/repo/users-repo.service.ts
@@ -306,6 +306,17 @@ export class UsersRepoService {
         });
     }
 
+    async deleteMapLibraryEntry(userID: number, mapID: number) {
+        await this.prisma.mapLibraryEntry.delete({
+            where: {
+                mapID_userID: {
+                    userID: userID,
+                    mapID: mapID
+                }
+            }
+        });
+    }
+
     //#endregion
 
     //#region Map Favorites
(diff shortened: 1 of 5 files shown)


server/src/modules/repo/users-repo.service.ts
```diff
@@ -306,6 +306,20 @@ export class UsersRepoService {
...
+            where: {
+                userID: userID,
+                mapID: mapID
+            }
+        });
+        await this.prisma.mapLibraryEntry.delete({
+            where: {
+                id: map.id
+            }
+        });
```
tsa96, line 320: Finding the map, *then* deleting it isn't ideal here, we should be able to do it in one go. The reason we can't do it currently is Prisma only lets you run a deletion on a *single* object, i.e. if it's guaranteed an entry is unique. We *could* use a `deleteMany` but that's ugly, the best practice is to write the uniqueness into the schema (MapNotify does this already btw).

So, in `server/prisma/schema.prisma` lets add the line 
```prisma
@@unique([mapID, userID])
```
to the model, below the `@@index`.

Then, we regenerate the prisma client with `npm run migrate` (if it errors, just `rm -rf prisma/migrations/`), and this should be allowed by the new prisma typings:
```ts
await this.prisma.mapLibraryEntry.delete({
    where: {
        mapID_userID: {
            userID: userID,
            mapID: mapID
        }
    }
});
```

server/src/modules/repo/users-repo.service.ts
```diff
@@ -330,6 +344,29 @@ export class UsersRepoService {
...
+    async createFavouritedMapEntry(userID: number, mapID: number) {
+        await this.prisma.mapFavorite.create({
+            data: {
+                userID: userID,
+                mapID: mapID
+            }
+        });
+    }
+
+    async deleteFavouritedMapEntry(userID: number, mapID: number) {
```
tsa96, line 356: Same thing with uniqueness as above comment

tsa96 (changes requested):
Everything here looks great, just the technical Prisma thing with the deletions

server/prisma/migrations/migration_lock.toml
```diff
@@ -0,0 +1,3 @@
+# Please do not edit this file manually
+# It should be added in your version-control system (i.e. Git)
+provider = "mysql"
```
tsa96, line 3: This file and `migration.sql` should not be committed! Please stash or just delete migrations if they're getting in the way of rebases or whatever

tsa96 (changes requested):

tsa96 (approved):

PR #683, opened 2023-01-01
✨ Admin - Get Maps

Hona:


server/src/modules/admin/admin.service.ts
```diff
@@ -185,4 +188,16 @@ export class AdminService {
...
+        // Where
+        const where: Prisma.MapWhereInput = {
+            status: status,
+            priority: priority
+        };
+
+        const dbResponse = await this.mapRepo.getAll(where, ExpandToPrismaIncludes(expand), null, skip, take);
+
+        return new PaginatedResponseDto(MapDto, dbResponse);
+    }
```
tsa96, line 202: I don't think we need this method. IMO the best approach is going to be to using the AdminCtlMapsGetAllQuery DTO, and MapsService.getAll can change to 
```ts
async getAll(userID: number, query: MapsCtlGetAllQuery | AdminCtlMapsGetAllQuery): Promise<PaginatedResponseDto<MapDto>> {
```
then add status and priority to the Prisma query if they're defined. Security-wise this should be fine, since class-validation will cause requests with bad queries to 400, so if a user calling GET maps/ included a status or priority, they'd 400.

tsa96: Worth writing an E2E test in maps.e2e-spec.ts to test that last bit!

server/src/modules/admin/admin.controller.ts
```diff
@@ -119,8 +119,8 @@ export class AdminController {
     @ApiOperation({ description: 'Retrieve a list of maps' })
     @ApiOkPaginatedResponse(MapDto, { description: 'Paginated list of maps' })
     @ApiBadRequestResponse({ description: 'Invalid query data' })
-    getMaps(@Query() _query: MapsCtlGetAllQuery): Promise<PaginatedResponseDto<MapDto>> {
-        return void 0;
+    getMaps(@Query() _query: AdminCtlMapsGetAllQuery): Promise<PaginatedResponseDto<MapDto>> {
+        return this.adminService.getMaps(_query.priority, _query.status, _query.skip, _query.take, _query.expand);
```
tsa96, line 123: Underscore prefixes in var names here were just to stop ESLint complaining, now they're actually being used they should be removed.

tsa96 (changes requested):
Needs some quite major restructures as per that comment on the service logic.

Also, I'm afraid E2E tests are mandatory for PRs, otherwise 1. we don't actually know if the code works and 2. it forces other people to write tests for you which is pretty demotivating for other contributors. I really can't stress enough how essential tests are (though I'm sure you're aware), this is one of the most significant quality improvements we're making over the old API and I want to make sure we stick with this attitude.

Last thing, sorry to be a nag but these commits really need to be cleaned up, the top 3 should be `fixup`ed into the below one. And please keep the commit naming scheme consistent with other commits in the repo (so feat:, chore: etc), I'm afraid most the team has a burning hatred of gitmojis

tsa96:
Closing this, inactive for almost two months and I'm reassigning the endpoint.

PR #688, opened 2023-01-08
Admin - GET, PUT xpsys

laees:


server/tests/e2e/admin.e2e-spec.ts
```diff
@@ -976,6 +1024,113 @@ describe('Admin', () => {
...
+            const res = await get({
+                url: 'admin/xpsys',
+                status: 200,
+                token: adminUserToken
+            });
+
+            expect(res.body).toBeValidDto(XpSystemsDto);
+            expect(res.body.cosXP.levels.maxLevels).toBe(600);
+        });
+
```
tsa96, line 1117: We should check that it *doesn't* allow moderators to use this.

server/src/common/dto/xp-systems/xp-systems.dto.ts
```diff
@@ -0,0 +1,171 @@
...
+    @ApiProperty()
+    staticScaleInterval: number;
+
+    @ApiProperty()
+    staticScaleIntervalMultiplier: number;
+}
+
+class CosXpDto {
+    [key: string]: Prisma.JsonObject;
+    @NestedDto(CosXpCompletions, { type: () => CosXpCompletions })
```
tsa96, line 83: Is the explicit `type: () => ...` needed here? That should only be required for avoiding circular references in Swagger. Applies to several of these files

server/src/common/dto/xp-systems/xp-systems.dto.ts
```diff
@@ -0,0 +1,171 @@
...
+import { NestedDto } from '@lib/dto.lib';
+import { IsNumber } from 'class-validator';
+import { Prisma } from '@prisma/client';
+
+// Could implement RankXpParams and CosXpParams in xp-systems.interface.ts but class-validator requires
+// we give it this wacky structure.
+
+// TODO: bunch of this can be readonly?
+class CosXpTierScale {
+    [key: string]: Prisma.JsonValue;
```
tsa96, line 11: I'm really confused by this, though I confess I don't think I've seen index signatures in classes before. I would assume it means "any string-based key must be a `Prisma.JsonValue`" but clearly we have values of other types in here. Can you explain to me what it's needed, either on here or Discord?

server/src/common/dto/xp-systems/xp-systems.dto.ts
```diff
@@ -0,0 +1,171 @@
...
+    @ApiProperty()
+    linear: number;
```
tsa96, line 13: Validators here please

tsa96 (changes requested):

tsa96 (approved):
Looks perfect, great work!

PR #691, opened 2023-01-13
feat: GET /api/maps/{mapID}/ranks/{rankNumber}

BossCode45:
For some reason the DB doesn't record trackNum or trackZone, so they are implemented but there are no tests for them and there are warnings in the DTO and a comment as a note.

### DIFF
--- a/server/src/common/dto/query/map-queries.dto.ts
+++ b/server/src/common/dto/query/map-queries.dto.ts
@@ -1,4 +1,6 @@
-﻿import { MapStatus, MapType } from '../../enums/map.enum';
+import { MapStatus, MapType } from '../../enums/map.enum';
+import { PaginationQuery } from './pagination.dto';
+
 import {
     BooleanQueryProperty,
     ExpandQueryProperty,
@@ -6,9 +8,9 @@ import {
     SkipQueryProperty,
     TakeQueryProperty,
     StringQueryProperty,
-    EnumQueryProperty
+    EnumQueryProperty,
+    IntCsvQueryProperty
 } from '@lib/dto.lib';
-
 class MapsGetAllBaseQuery {
     @SkipQueryProperty(0)
     skip = 0;
@@ -80,4 +82,29 @@ export class MapsGetQuery {
 export class MapCreditsGetQuery {
     @ExpandQueryProperty(['user'])
     expand: string[];
+}
+
+export class MapRanksGetQuery extends PaginationQuery {
+    @IntQueryProperty({ description: 'Steam ID of player to get rank for' })
+    playerID: number;
+
+    @IntCsvQueryProperty({ description: 'CSV list of steam IDs of players to get rank for' })
+    playerIDs: number[];
+
+    @IntQueryProperty({ description: 'Rank flags', default: 0 })
+    flags: number;
+
+    @BooleanQueryProperty({ description: 'Whether to order by date or not (false for reverse)' })
+    orderByDate: boolean;
+}
+
+export class MapRankGetNumberQuery {
+    @IntQueryProperty({ description: 'Track number', default: 0 })
+    trackNum: number;
+
+    @IntQueryProperty({ description: 'Zone number', default: 0 })
+    zoneNum: number;
+
+    @IntQueryProperty({ description: 'Rank flags', default: 0 })
+    flags: number;
 }
--- a/server/src/modules/maps/maps.service.ts
+++ b/server/src/modules/maps/maps.service.ts
@@ -6,8 +6,9 @@ import {
     Injectable,
     NotFoundException
 } from '@nestjs/common';
-import { Map as MapDB, MapCredit, MapTrack, Prisma } from '@prisma/client';
+import { Map as MapDB, MapCredit, MapTrack, Prisma, UserMapRank } from '@prisma/client';
 import { CreateMapDto, MapDto, UpdateMapDto } from '@common/dto/map/map.dto';
+import { MapRankDto } from '@common/dto/map/map-rank.dto';
 import { PaginatedResponseDto } from '@common/dto/paginated-response.dto';
 import { MapsRepoService } from '../repo/maps-repo.service';
 import { AuthService } from '../auth/auth.service';
@@ -19,8 +20,9 @@ import { ActivityTypes } from '@common/enums/activity.enum';
 import { CreateMapCreditDto, MapCreditDto, UpdateMapCreditDto } from '@common/dto/map/map-credit.dto';
 import { MapInfoDto, UpdateMapInfoDto } from '@common/dto/map/map-info.dto';
 import { MapTrackDto } from '@common/dto/map/map-track.dto';
-import { MapsCtlGetAllQuery } from '@common/dto/query/map-queries.dto';
-import { MapImageDto } from '../../common/dto/map/map-image.dto';
+import { MapsCtlGetAllQuery, MapRanksGetQuery, MapRankGetNumberQuery } from '@common/dto/query/map-queries.dto';
+import { UsersService } from '../users/users.service';
+import { MapImageDto } from '@common/dto/map/map-image.dto';
 import { FileStoreCloudFile } from '../filestore/file-store.interfaces';
 import { ConfigService } from '@nestjs/config';
 import sharp from 'sharp';
@@ -34,7 +36,8 @@ export class MapsService {
         private readonly mapRepo: MapsRepoService,
         private readonly userRepo: UsersRepoService,
         private readonly fileCloudService: FileStoreCloudService,
-        private readonly config: ConfigService
+        private readonly config: ConfigService,
+        private readonly usersService: UsersService
     ) {}
 
     //#region Maps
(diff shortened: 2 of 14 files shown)


server/src/common/dto/query/map-queries.dto.ts
```diff
@@ -176,3 +176,35 @@ export class MapRanksGetQuery extends PaginationQuery {
     @IsOptional()
     orderByDate: boolean;
 }
+
+export class MapRankGetNumberQuery {
+    @ApiPropertyOptional({
+        name: 'trackNum',
+        type: Number,
+        description: 'Track number (default 0)'
```
tsa96, line 184: You can use the `default` property here so `default: 0` which swagger will show, better than typing the default manually

tsa96: Applies in a couple cases below

server/tests/e2e/maps.e2e-spec.ts
```diff
@@ -2396,4 +2458,93 @@ describe('Maps', () => {
...
+            const res = await get({
+                url: `maps/${map1.id}/ranks/1`,
+                status: 200,
+                token: user1Token
+            });
+
+            expect(res.body.rank).toBe(1);
+            expect(res.body.mapID).toBe(map1.id);
+            expect(res.body.userID).toBe(admin.id);
+            expect(Number.parseInt(res.body.runID)).toBe(Number.parseInt(run1.id));
```
tsa96, line 2473: This could pass incorrectly for large BigInts, since they lose precision when converted to JSON. For example
`Number.parseInt(BigInt(111111111111111111111))` becomes `111111111111111110000`. So use `BigInt(x)`, not `Number.parseInt(x)`

server/src/common/dto/query/map-queries.dto.ts
```diff
@@ -133,3 +134,45 @@ export class MapCreditsGetQuery {
...
+    playerID: number;
+
+    @ApiPropertyOptional({
+        name: 'playerIDs',
+        type: Number,
+        isArray: true,
+        description: 'List of steam IDs of players to get rank for'
+    })
+    @IsArray()
+    @Transform(({ value }) => value.map((v) => Number.parseInt(v)))
```
tsa96, line 156: I think you can just use `Type()` here since class-transformer will figure out you're transforming an array https://github.com/typestack/class-transformer#working-with-arrays

tsa96: But wait, this doesn't even make sense. This is on a query, which doesn't support arrays; it's just plaintext. So you want to pass a string and split it, like here (L48 user-queries.dto.ts)
```ts
@ApiPropertyOptional({
    name: 'userIDs',
    type: String,
    description: 'Filter by CSV list of Steam Community IDs',
    example: '123135674,7987347263,98312287631'
})
@Transform(({ value }) => value.split(','))
@IsSteamCommunityID({ each: true })
@IsOptional()
steamIDs: string[];
```

server/src/common/dto/query/map-queries.dto.ts
```diff
@@ -133,3 +134,45 @@ export class MapCreditsGetQuery {
...
+    @IsOptional()
+    playerIDs: number[];
+
+    @ApiPropertyOptional({
+        name: 'flags',
+        type: Number,
+        description: 'Rank flags (default 0)'
```
tsa96, line 163: Another weird default

server/src/modules/maps/maps.service.ts
```diff
@@ -440,6 +441,29 @@ export class MapsService {
...
+        const where: Prisma.UserMapRankWhereInput = {
+            mapID: mapID,
+            flags: query.flags
+        };
+
+        if (query.playerID) where.userID = query.playerID;
+        if (query.playerIDs) where.userID = { in: query.playerIDs };
+
+        const order: Prisma.UserMapRankOrderByWithAggregationInput = {};
+        if (query.orderByDate !== undefined) {
```
tsa96, line 458: You don't need to check for `undefined` specifically here, validation will have sanitised this. Just ` if (query.orderByDate)` is fine.

server/tests/e2e/maps.e2e-spec.ts
```diff
@@ -518,6 +520,35 @@ describe('Maps', () => {
...
+            data: {
+                map: { connect: { id: map1.id } },
+                user: { connect: { id: user2.id } },
+                trackNum: 1,
+                zoneNum: 1,
+                ticks: 200000,
+                tickRate: 100,
+                flags: 123,
+                file: '',
+                hash: '35a3a113d817ad904d471ffffb6a966cb241a350',
```
tsa96, line 535: Just randomly searched the existing code for this exact hash and it is indeed being used somewhere else. Please don't just duplicate data that's obviously supposed to be random, it's bound to end up breaking unique constraints in the DB at some point.

tsa96: I'm not going to check every test in this PR, please look through and make sure everything is unique yourself

server/tests/e2e/maps.e2e-spec.ts
```diff
@@ -2225,4 +2256,144 @@ describe('Maps', () => {
...
+                url: `maps/${map1.id}/ranks`,
+                status: 200,
+                token: user1Token
+            });
+
+            expect(ranks.body.totalCount).toBe(3);
+            expect(ranks.body.returnCount).toBe(3);
+            expect(ranks.body.response[0].mapID).toBe(map1.id);
+            expect(ranks.body.response[0].userID).toBe(user1.id);
+            expect(Number.parseInt(ranks.body.response[0].runID)).toBe(Number.parseInt(run2.id));
```
tsa96, line 2334: Another case where we should use `BigInt`s not `Number`s

server/tests/e2e/maps.e2e-spec.ts
```diff
@@ -2225,4 +2256,144 @@ describe('Maps', () => {
...
+                query: {
+                    playerID: user1.id
+                }
+            });
+
+            expect(ranks.body.totalCount).toBe(1);
+            expect(ranks.body.returnCount).toBe(1);
+            expect(ranks.body.response[0].mapID).toBe(map1.id);
+            expect(ranks.body.response[0].userID).toBe(user1.id);
+            expect(Number.parseInt(ranks.body.response[0].runID)).toBe(Number.parseInt(run2.id));
```
tsa96, line 2357: bigint

server/tests/e2e/maps.e2e-spec.ts
```diff
@@ -2225,4 +2256,144 @@ describe('Maps', () => {
...
+            expect(ranks.body.response[0].rankXP).toBe(0);
+        });
+
+        it('should return only runs for a set of players when given the query param playerIDs', async () => {
+            const ranks = await get({
+                url: `maps/${map1.id}/ranks`,
+                status: 200,
+                token: user1Token,
+                query: {
+                    playerIDs: [admin.id, user2.id]
```
tsa96, line 2372: This is gonna need to be changed to a string like
 ```ts
playerIDs: `{admin.id},{user2.id}`
```

**Review by tsa96 — CHANGES_REQUESTED** [review]
Bunch of comments but it's a pretty large PR, overall you did the refactor really well. One other thing to change though, your UMR refactor PR is marked `feat: ` when it should definitely be `refactor: `.

File: server/src/common/dto/query/map-queries.dto.ts (hunk 3 above)
Comment (BossCode45) [reply server/src/common/dto/query/map-queries.dto.ts:156]: queries do actually support arrays
`?userIDs=1234&userIDs=5678` should make `"userIDs": ["1234", "5678"]`

File: server/tests/e2e/maps.e2e-spec.ts (hunk 6 above)
Comment (BossCode45) [reply server/tests/e2e/maps.e2e-spec.ts:535]: How should i generate the hashes?

Comment (tsa96) [reply server/tests/e2e/maps.e2e-spec.ts:535]: Oh good question, these should be genuine SHA1 hashes. Easiest way is to just use an online tool like [here](https://md5decrypt.net/en/Sha1). Interestingly enough, the hash here decodes to the word "ostriches" :)

**Comment by BossCode45** [comment]
There is no testing for `GET /api/maps/{rankNumber}/ranks/friends` since no users have  actual steamIDs

File: server/src/modules/maps/maps.service.ts [hunk 10]
```diff
@@ -440,6 +449,226 @@ export class MapsService {
...
+        );
+
+        return ret;
+    }
+
+    async getRankFriends(userID: number, mapID: number, query: MapRankGetNumberQuery): Promise<MapRankDto[]> {
+        const user = await this.userRepo.get(userID);
+        const steamID = user.steamID;
+
+        const res = await firstValueFrom(
```
tsa96, line 596: Query is mostly right but I think we should split this up into its own function in `users.service.ts`. You'll need to update `maps.module.ts` to import the `UsersService`, I don't mind doing that.

Then let's type up the response from Steam, you can add that to `auth.interface.ts` for now. Maybe not the best place for that file, in the future we can move that out to something like `lib/steam/interfaces.ts`.

server/src/modules/maps/maps.service.ts
```diff
@@ -440,6 +449,226 @@ export class MapsService {
...
+                        steamID: steamID,
+                        relationship: 'friend'
+                    }
+                })
+                .pipe(
+                    catchError((error: AxiosError) => {
+                        if (error.response.status === 401)
+                            throw new ConflictException('Friends list or profile is private');
+
+                        throw new BadRequestException('Bad request');
```
tsa96, line 610: If we don't get a response from Steam, that's really a problem on our end, not the client. Remember, 4xxs are *client* errors such as bad request data or attempting to access a resource that doesn't exist.

We're responsible for the services our system uses, including Steam, so if there's something wrong in their end (classic Tuesday downtime, thanks Valve), so lets throw a `ServiceUnavailableException('Failed to retrieve friends list from Steam');` or whatever Nest calls its 503 exceptions.

server/src/modules/maps/maps.service.ts
```diff
@@ -440,6 +449,226 @@ export class MapsService {
...
+
+        if (res.data) {
+            const friendSteamIDs = res.data.friendslist.friends.map((item) => item.steamid);
+
+            if (friendSteamIDs.length === 0) throw new ImATeapotException('No friends detected :(');
+
+            //This must be done like this because mapping promises is wierd
+            const friendPromises = friendSteamIDs.map((item) => this.userRepo.getBySteamID(item));
+            const friends: User[] = await Promise.all(friendPromises);
+            const friendIDs = friends.map((item) => item.id);
```
tsa96, line 623: Funny comment here, this is a very normal use of promises in JS. This is the standard way of parallelising a bunch of stuff, i.e. make an array of promises and await them all - `Promise.all` returns a new promise that runs an array of promises in parallel and resolves when they've all resolved.

That being said, this isn't the right way to do this. Rather than make a bunch of separate DB queries ourselves, it's best to just let Prisma handle it for us, as there's likely optimisations made that'll make it run faster.

So instead, we can do something like `const friends = await this.userRepo.getAll({ steamID: { in: friendSteamIDs } }, {});` (for some reason getAll requires an `include` object, IMO would be worth changing that, just making the include optional. but commit separately if you do that)

tsa96: Actually, why are you doing this at all? SteamIDs are unique to `User`s in the database, just use that in your `where`. So just scrap all this and change your getRanks `where` to use `steamID`s instead of `userID`s.

server/tests/e2e/maps.e2e-spec.ts
```diff
@@ -2458,4 +2458,178 @@ describe('Maps', () => {
...
+                            }
+                        },
+                        overallStats: {
+                            create: {
+                                jumps: 1
+                            }
+                        }
+                    }
+                });
+            }
```
tsa96, line 2569: This is a nit but I prefer we use arrow functions for stuff like this.

server/tests/e2e/maps.e2e-spec.ts
```diff
@@ -2458,4 +2458,178 @@ describe('Maps', () => {
...
+            });
+
+            let prevRank = 1;
+            for (const rank of res.body) {
+                expect(rank).toBeValidDto(MapRankDto);
+                expect(rank.rank).toBe(prevRank + 1);
+                prevRank = rank.rank;
+            }
+            expect(res.body.length).toBe(11);
+        });
```
tsa96, line 2606: Please clean up all the users and runs you created at the end. In fact when we're doing DB setup for individual tests, we should probably ensure teardown happens even if the test fails. Apparently `try..finally` works (https://stackoverflow.com/a/62958360/19578983) so lets wrap all this in a `try` and in a `finally` block, teardown everything you added here (catch and ignore any failed deletions).

server/tests/e2e/maps.e2e-spec.ts
tsa96: Also please at least insert newlines between function declarations!

server/tests/e2e/maps.e2e-spec.ts
```diff
@@ -2458,4 +2458,178 @@ describe('Maps', () => {
                 status: 401
             }));
     });
+
+    describe('GET /api/maps/{mapID}/ranks/{rankNumber}', () => {
```
tsa96, line 2462: Even if on a certain level ranks/{blah} is the same endpoint (eh, it really isn't), lets split out the three variants into separate `describe` blocks.

tsa96: Or I guess two variants since we're not testing /friends. I've made a note to unit test that in the future.

server/src/modules/maps/maps.service.ts
```diff
@@ -440,6 +449,226 @@ export class MapsService {
...
+        if (query.playerIDs) where.userID = { in: query.playerIDs };
+
+        const include = {
+            run: true
+        };
+
+        const order: Prisma.UserMapRankOrderByWithAggregationInput = {};
+        if (query.orderByDate) {
+            order.createdAt = query.orderByDate ? 'desc' : 'asc';
+        }
```
tsa96, line 472: Nit I know, but no need to braces here

tsa96 (changes requested):

server/src/modules/maps/maps.service.ts
tsa96: Make sure you place this in a good location in the users.service file. Change the "Private" region to "Steam Queries" and add there (public of course, leave rest private until they're needed elsewhere)

tsa96 (approved):

PR #692, opened 2023-01-22
PATCH maps/:mapID

laees:


server/src/modules/maps/maps.service.ts
```diff
@@ -248,6 +248,39 @@ export class MapsService {
         return mapDB.id;
     }
 
+    async update(mapID: number, userID: number, updateBody: UpdateMapDto): Promise<void> {
+        const map = await this.mapRepo.get(mapID);
```
laees, line 252: I don't know if there are status flags that not every user is allowed to set. If there are then a check needs to be added here. 
As far as I could see the old API doesn't check for it so I didn't either

server/src/modules/maps/maps.controller.ts
```diff
@@ -105,6 +105,27 @@ export class MapsController {
         return this.mapsService.get(mapID, userID, query.expand);
     }
 
+    @Patch('/:mapID')
+    @HttpCode(HttpStatus.NO_CONTENT)
+    @Roles(RolesEnum.MAPPER)
+    @ApiOperation({ summary: 'Updates a single maps status flag' })
```
tsa96, line 111: Map's (use double quotes for whole string)

server/src/modules/maps/maps.service.ts
tsa96: Yeah the permissions system here does seem a bit odd, let's just implement it the same way for now.

Though one thing you're missing, the old API is changing that the map is not in the `REJECTED` or `REMOVED` state, let's at least add that. So just return a 400 if it's in either of those states.

tsa96 (changes requested):
Awesome stuff, very few issues. You handled the activities creation and testing very succinctly, so much nicer than old API!

tsa96 (approved):

PR #697, opened 2023-02-05
feat: Added types and descriptions to ApiOKResponses

GordiNoki:


tsa96 (approved):

tsa96:
Have you run Prettier? Should be a commit hook for it, if not I'll check that out. Not sure what it's complaining about though tbh.

GordiNoki:
Neither do I, nest build runs fine, but looks like prettier doesn't like nested types there. I'll remove angle brackets then

PR #698, opened 2023-02-07
feat: Unified logging style to Logger instances

GordiNoki:


GordiNoki:
Ok, commit hook definitely doesn't work when committing with VSC

tsa96 (approved):

PR #699, opened 2023-02-18
GET maps/:mapID/download

laees:
There are some things I'd like to get input about, which I was very unsure of while writing the changes

1.  I didn't know if I should log a failed download, I decided to do so and include the error message, since an invalid key or bucket can  throw different errors
1.  I included the response type from express to pass the object to the service, I think I could have gotten away with the `any` type but did so for clarity
1.  If `maps.service.ts` LN: 330 fails, the download won't start. I thought about catching and ignoring the potential error there so that a failed downloadCounter increment doesn't stop a download, alternatively I could have created a service function to increment the counter after writing to the response

### DIFF
--- a/server/src/modules/filestore/file-store-cloud.service.ts
+++ b/server/src/modules/filestore/file-store-cloud.service.ts
@@ -1,6 +1,6 @@
-import { Injectable, Logger } from '@nestjs/common';
+import { Injectable, Logger, StreamableFile } from '@nestjs/common';
 import { FileStoreCloudFile } from './file-store.interfaces';
-import { S3Client, PutObjectCommand, DeleteObjectCommand } from '@aws-sdk/client-s3';
+import { S3Client, PutObjectCommand, DeleteObjectCommand, GetObjectCommand } from '@aws-sdk/client-s3';
 import { FileStoreUtils } from './file-store.utility';
 import { ConfigService } from '@nestjs/config';
 
@@ -48,5 +48,24 @@ export class FileStoreCloudService {
         );
 
         this.logger.log(`DELETE SUCCESS! Deleted file ${fileKey} from bucket ${this.config.get('storage.bucketName')}`);
+    }
+
+    async getFileCloud(fileKey: string): Promise<StreamableFile> {
+        try {
+            const cmdRes = await this.s3Client.send(
+                new GetObjectCommand({
+                    Bucket: this.config.get('storage.bucketName'),
+                    Key: fileKey
+                })
+            );
+
+            return new StreamableFile(await cmdRes.Body.transformToByteArray());
+        } catch (error) {
+            this.logger.error(
+                `DOWNLOAD FAILED! Err:"${error.message}" Download file ${fileKey} from bucket ${this.config.get(
+                    'storage.bucketName'
+                )}`
+            );
+        }
     }
 }
--- a/server/src/modules/maps/maps.controller.ts
+++ b/server/src/modules/maps/maps.controller.ts
@@ -15,7 +15,9 @@ import {
     Patch,
     Delete,
     Put,
-    UseGuards
+    UseGuards,
+    Header,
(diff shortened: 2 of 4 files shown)


server/src/modules/maps/maps.service.ts
```diff
@@ -317,6 +318,33 @@ export class MapsService {
...
     }
 
+    async download(mapID: number, res: Response): Promise<void> {
+        const map = await this.mapRepo.get(mapID);
+
+        if (!map) throw new NotFoundException('Map not found');
+
+        const mapData = await this.getMapFileFromStore(map.name);
+
+        if (!mapData) throw new NotFoundException(`Couldn't find bsp file for ${map.name}.bsp`);
```
tsa96, line 328: "BSP"

server/src/modules/maps/maps.service.ts
```diff
@@ -317,6 +318,33 @@ export class MapsService {
...
+        await this.mapRepo.updateMapStats(mapID, {
+            downloads: { increment: 1 }
+        });
+
+        res.set({
+            'Content-Type': 'application/octet-stream',
+            'Content-Length': mapData.length
+        });
+        res.write(mapData, 'binary');
+        res.end();
```
tsa96, line 339: Services should handle requests, that's the controller's job. So don't pass the Response to the service, have the service response a UIntArray

server/src/modules/filestore/file-store-cloud.service.ts
```diff
@@ -49,4 +49,23 @@ export class FileStoreCloudService {
...
+    async getFileCloud(fileKey: string): Promise<Uint8Array> {
+        try {
+            const cmdRes = await this.s3Client.send(
+                new GetObjectCommand({
+                    Bucket: this.config.get('storage.bucketName'),
+                    Key: fileKey
+                })
+            );
+
+            return cmdRes.Body.transformToByteArray();
```
tsa96, line 62: Best method (https://docs.nestjs.com/techniques/streaming-files) seems to be to return this as Nest's StreamablePipe which handles piping for us. So
```ts
async getFileCloud(fileKey: string): Promise<StreamableFile> {
    try {
        const cmdRes = await this.s3Client.send(
            new GetObjectCommand({
                Bucket: this.config.get('storage.bucketName'),
                Key: fileKey
            })
        );

        return new StreamableFile(await cmdRes.Body.transformToByteArray());
```

server/src/modules/maps/maps.controller.ts
```diff
@@ -199,6 +201,22 @@ export class MapsController {
...
+    @ApiOkResponse({ description: "The map's bsp file" })
+    @ApiNotFoundResponse({ description: 'Map was not found' })
+    @ApiNotFoundResponse({ description: "Map's bsp file could not be found" })
+    @ApiParam({
+        name: 'mapID',
+        type: Number,
+        description: 'Target Map ID',
+        required: true
+    })
+    downloadMap(@Res({ passthrough: true }) res: Response, @Param('mapID', ParseIntPipe) mapID: number) {
```
tsa96, line 216: Explicitly put the return type here, i.e. `StreamableFile`

server/src/modules/maps/maps.service.ts
tsa96: Reading up on Nest docs, we don't need to do all this anyway. Once the file storage stuff is changed to return a `StreamableFile`, we just pass that back to the controller.

Note: we're going to change this whole approach anyway to just send a bucket URL to the client. But worth implementing that file store stuff properly for any files we do end up streaming through the backend.

tsa96 (changes requested):
Sorry I took so long to get to this, should've done last week. The file handling stuff isn't quite right, but close, and good job working with the S3 stuff!

tsa96 (approved):
Great stuff, thank you!

PR #703, opened 2023-02-25
feat: added /admin/maps endpoints logic

GordiNoki:


server/src/modules/runs/runs.service.ts
```diff
@@ -64,4 +65,18 @@ export class RunsService {
 
         return new PaginatedResponseDto(RunDto, dbResponse);
     }
+
+    async deleteStoredMapRuns(mapID: number): Promise<void> {
+        const where: Prisma.RunWhereInput = {
+            mapID
+        };
+        const dbResponse = await this.runRepo.getAllRuns(where);
```
tsa96, line 73: Nit: Just do `const dbResponse = await this.runRepo.getAllRuns({ mapID });`.

server/src/modules/runs/runs.service.ts
```diff
@@ -64,4 +65,18 @@ export class RunsService {
...
+        const runs = dbResponse.flatMap<Run>((resp) => resp[0]);
```
tsa96, line 74: I'm really confused by what you're doing here. `getAllRuns` returns a `[Run[], number]` tuple, so mapping the elements of the tuple to some subelement is going to break for the second value (the count). Out of curiosity I checked, and in JS trying to access the index of a number just returns `undefined`, regardless it's certainly not something we should be doing. It will definitely fail tests once they're written.

You just want the arrays of runs, so you only need `dbResponse[0]`. I would probably just do `const [runs] = await this.runRepo.getAllRuns(where);` since `(await this.runRepo.getAllRuns(where))[0];` looks rather unpleasant.

server/src/modules/maps/maps.service.ts
```diff
@@ -291,6 +302,21 @@ export class MapsService {
...
     }
 
+    async delete(mapID: number): Promise<void> {
+        const map = await this.mapRepo.get(mapID);
+        if (!map) throw new NotFoundException('No map found');
+
+        // Deleting all stored map images
+        const imageDtos = await this.getImages(mapID);
+        const imageFileDeletes = imageDtos.map((dto) => this.deleteStoredMapImage(dto.id));
+        await Promise.all(imageFileDeletes);
```
tsa96, line 312: Kinda pointless assignment here, I would just do `await Promise.all(imageDtos.map((dto) => this.deleteStoredMapImage(dto.id)));`.

server/src/modules/runs/runs.service.ts
```diff
@@ -64,4 +65,18 @@ export class RunsService {
...
+
+        const runFileDeletes = runs.map((run) => {
+            this.fileCloudService.deleteFileCloud(`runs/${run.id}`);
+        });
```
tsa96, line 78: Unnecessary arrow func braces

tsa96: Again a weird assignment, I was just promise.all the mapping immediately. Not a big deal though.

tsa96 (changes requested):
Most code is okay besides the weird thing in the stored run deletion.

Main thing though, this needs tests. No endpoint is getting merged in with E2E tests unless there's a very good reason, testing is absolutely essential for web API development and should be taken as seriously as the rest of the codebase. I get that tests can be pretty boring, but it's better than dealing with dozens of random bugs when in production.

Plus, it doesn't look like the run file deletion is even going to work, writing tests would've shown that! See existing test files for general approach on how we write E2E tests, then these should be pretty simple. The only complex thing is, we probably want to check that map image files actually have been deleted from the S3 bucket, so you'll need to set up as instance in the test file (I think maps e2e does this) and do a check with the S3 API

tsa96:
Comments are all resolved, but please `fixup` those changes into your previous commit and force-push - we don't like include `fix: ` comments that are addressing issues brought up in the PR, since once this branch is merged, that fix commit isn't a helpful addition to the Git commit history (admittedly there are a *lot* of fix commits on this repo, but that's cus I'm constantly refactor and fixing stuff as the rewrite moves forward, hopefully far less once tests are refactored and hooked up to CI).

Also, looks like there's some merge conflicts to resolve now other branches have been merged, let me know if you need a hand resolving those.

tsa96:
Oh also, just so we can easily search git history to say precise what the added endpoints are, so `reword` to `feat: /admin/maps GET, PATCH, DELETE`

tsa96 (changes requested):
Everything here is good, just waiting on tests, see Discord comments

tsa96 (changes requested):
Started reviewing this then had a test fail when I checked something, I'm going to just clean this up myself real quick and get it in. Almost everything you've done is excellent though, tests look really solid.

tsa96 (approved):

PR #712, opened 2023-03-19
Fix API versioning

benjl:
`auth.*` doesn't get parsed as a regular expression when trying to exclude paths from `setGlobalPrefix`, and instead uses wildcards provided by path-to-regexp. `auth(.*)` is parsed as a regular expression so it properly excludes auth. Hooray


tsa96 (approved):
💕

PR #717, opened 2023-03-19
refactor: updated the runs tests

BossCode45:
A comment beginning with `LEGACY` marks legacy code (from before the refactor) that I've kept for some reason, reason explained in the comment, the code under could probably change, but I've left it just in case it breaks stuff (so I can get the PR out fast).
A comment beginning with `CHECK` is where something weird is going on, and it requires an extra look over, this code should probably change, I just don't know how.


tsa96:
Sorry but this isn't needed, I'm going to just rebase these changes into the overall tests refactor PR, way easier.

PR #723, opened 2023-03-21
Hide REPORT_FILED activities from users/:userID/activities

benjl:


server/tests/e2e/users.e2e-spec.ts
```diff
@@ -291,6 +291,19 @@ describe('Users', () => {
                 token: token
             }));
 
+        it('should not include REPORT_FILED activities', async () => {
+            prisma.activity.create({ data: { type: ActivityTypes.REPORT_FILED, data: 119n, userID: user.id } });
```
tsa96, line 295: Missing await! Surprised this passes the validatePaged count tbh...

tsa96 (changes requested):
I'm fast

server/tests/e2e/users.e2e-spec.ts
benjl: oops fixed!!

PR #727, opened 2023-03-25
Feat/postgres

murgle2:
Notes

- Docker config now uses default postgres port; removed port from config
- Docker db config now uses docker compose's network, meaning db url uses name of service (@db instead of @localhost)
- Location of DB volume changed from /var/lib/mysql to ../postgres:/var/lib/postgresql/data
- DB initialized with only a singer user now
- PG doesn't support unsigned numeric datatypes. Implication is that max value of numeric fields will change despite using equivalent datatype (e.g., MySQL's unsigned int has a max value of 4294967295 whereas PG's int has a max value of 2147483647)


tsa96 (changes requested):
This is looking good now, though you have some redundant commits now, making changes just addressing changes made in your PR. Please `fixup`!

tsa96:
I'll tackle this once I have time, we need to get all tests passing first, was definitely something broken with cascade deletions. Let me know if you get a chance to tackle this, otherwise I'll get to it once Fastify branch is finished.

tsa96 (approved):

Issue #748, opened 2023-06-05
Optimize size of frontend webpack bundling


tsa96:
Our currently bundle size is sitting at around 4MB, which is unpleasantly large for a site of this size. I don't have time to focus on this right now, but I've installed webpack-bundle-analyzer and added a Nx command - do a prod build then run `nx run frontend:bundle-analzyer`. Here's a screenshot of it currently:

![Webpack_Bundle_Analyzer_ 5_Jun_2023_at_1850 _-_Goo05,06,23,185041](https://github.com/momentum-mod/website/assets/35813309/4774f57f-6302-443d-88b5-35b2704722ce)

A couple things we can likely improve very easily
- I think we can nuke Eva Icons. The app will spit out browser errors without it currently, but we can hopefully set Material Design Icons as the default pack (see https://akveo.github.io/nebular/docs/components/icon/api#nbiconlibraries), unless there's some Nebular components that use them internally. Even if there are, I don't know why the build process isn't tree-shaking the unused ones, as it does for MDI and Simple-Icons
- It looks we're shipping duplicates of Marked and ECharts in scripts.js and within custom packages like NgxEcharts that include them anyway. We can potentially get rid of Marked anyway, since we only use it for make descriptions and profile bios, which we show ingame as well, which doesn't have a markdown parser.

I may have also just done some imports badly and be duplicating modules there, potentially including the same stuff in separate modules that are lazy-loaded. Only been working with Angular seriously for the past couple months so likely to be making some mistakes.

tsa96:
Okay the `staging` version is only like 1MB, clearly I've screwed something up.

tsa96:
As Nick mentioned in https://github.com/momentum-mod/website/pull/771, we bumped the `maximumError` value in `apps/frontend/projects.json` so prod builds pass for now, remember to decrease back down during this.

tsa96:
Angular 16 supports esbuild and other more modern bits of tooling, though `nx-angular` doesn't seem to yet. Couldn't find any info on that, we can maybe move to using an Angular-based executor rather than Nx (`@nx/angular` is mostly just a wrapper around the Angular version I think)

Let's get everything (#775) to esbuild first, then work on the frontend optimization.

PR #771, opened 2023-06-26
Nx CI

tsa96:
Adds CI jobs for pull requests. Makes use of the glorious free runners from Github and runs the three biggest ones in separate jobs, really happy with the speed so far.

The frontend-e2e one isn't particularly useful yet, we literally only have one test, however it does verify that the site loads in a real browser without errors. And we'll hopefully be adding lots more frontend e2e testing in the future.

### DIFF
--- a/apps/frontend/project.json
+++ b/apps/frontend/project.json
@@ -43,7 +43,7 @@
             {
               "type": "initial",
               "maximumWarning": "500kb",
-              "maximumError": "1mb"
+              "maximumError": "5mb"
             },
             {
               "type": "anyComponentStyle",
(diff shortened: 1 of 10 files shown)


apps/frontend/project.json
```diff
@@ -43,7 +43,7 @@
             {
               "type": "initial",
               "maximumWarning": "500kb",
-              "maximumError": "1mb"
+              "maximumError": "5mb"
```
Gocnak, line 46: Would be good to make note to change this in #748

Gocnak (approved):
Nicely done! CI is an exciting gift that keeps on giving

Issue #767, opened 2023-06-24
Allow users to update their country


tsa96:
People often use silly countries on Steam, plus we only check this on first registration.

Supporting this is really simple:
1. Add a property for this in the /user PATCH endpoint. Validate using class-validator's @IsISOwhateveritscalledCode.
2. Add E2E tests for above
3. Add a dropdown to the frontend where users can pick any country - we have some flag icons in the frontend already, use those. Will need a list of countries and their country codes from somewhere. Grab those and I guess put in an object/map in `lib/constants`?

Great first issue for someone looking for some simple work on both front and backend.

tsa96:
Closed by https://github.com/momentum-mod/website/pull/772

PR #772, opened 2023-06-28
Country select and flags fix

RogiJAG:
PR addressing #767 - Allowing users to update their country. Also fixes country flag display.

Changes:
Adds field `countryCode` to the api/v1/users PATCH endpoint.
Adds backend e2e tests for the api/v1/users PATCH endpoint.
Adds country selector to edit profile page.
Fixes country flag display on user profile page (previously broken by update to package "flag-icons").
Adds ISOCountryCode enum - lists all country codes and the country's displayable name.


tsa96 (approved):
Amazing work! I don't have any issues with the code, you've done everything perfectly.

Only issue is commit messages, you've got stuff like `feat: do something in backend` where it's easier to read and cleaner to do `feat(back): do something`. I'll have CI checks hooked up with commitlint for this really soon so Github will tell you automatically. But since this is such a minor thing and I'm happy merging this anyway, I renamed myself and force-pushed myself (normally wouldn't do this but like I said, branch is getting merged right now anyway)

Issue #744, opened 2023-05-22
Rewrite Community News component to use new blog


tsa96:
We currently have a component called `community-news` that uses the Tumblr to get content from our old Tumblr blog. Fortunately, we're no longer using Tumblr (thank god), [the new one](https://github.com/momentum-mod/blog) is a static site using Hugo and is already configured to generate an RSS feed we can use.

I've no experience with RSS whatsoever, but looks easy. XML feed is here at https://blog.momentum-mod.org/index.xml, will probably need to install `xml2js` or similar unless there's some native browser method I'm unaware of.

tsa96:
Note: blocked until Nx branch is in.

PR #776, opened 2023-07-02
refactor(front): change community-news from Tumblr blog to new blog

RogiJAG:
PR addressing #744  - Use the new blog feed in the community-news component rather than the old Tumblr blog feed.

Changes:
Refactors TumblrApiService to BlogService.
Creates community-news component to use new blog service.
Adds style rules to override some of the markdown formatting from the blog.
Adds txml dependency.

Notes:
Does not fix unit tests for the blog service (it seems most of the unit tests are broken so I didn't touch them).


apps/frontend/src/app/pages/dashboard/community/community-news/community-news.component.ts
```diff
@@ -1,8 +1,55 @@
...
+      });
+  }
+}
+
+interface BlogPost {
+  guid: any;
+  title: any;
+  body: any;
+  pubDate: any;
+  link: any;
```
tsa96, line 54: Don't we know the type of these? Aren't they all just `string`s? If so use that

apps/frontend/src/app/pages/dashboard/community/community-news/community-news.component.ts
```diff
@@ -1,8 +1,55 @@
...
-  styleUrls: ['./community-news.component.scss']
+  styleUrls: ['./community-news.component.scss'],
+  encapsulation: ViewEncapsulation.None
 })
-export class CommunityNewsComponent {}
+export class CommunityNewsComponent implements OnInit {
+  // The number of blog posts to display on this component
+  private readonly POSTS_DISPLAYED = 10;
+
+  feed: any;
```
tsa96, line 16: May as well type this up a bit (though we don't need to care about the other crap besides `<item>`s. Looks like could be 
```ts
this.feed: { item: BlogPost[]; } = {};
```

libs/frontend/data/src/lib/blog.service.ts
```diff
@@ -0,0 +1,14 @@
+import { Injectable } from '@angular/core';
+import { HttpClient } from '@angular/common/http';
+import { Observable } from 'rxjs';
+
+@Injectable({ providedIn: 'root' })
+export class MomBlogService {
```
tsa96, line 6: "Mom" prefix feels unnecessary here!

tsa96 (changes requested):

apps/frontend/src/app/pages/dashboard/community/community-news/community-news.component.ts
```diff
@@ -1,8 +1,55 @@
-import { Component } from '@angular/core';
+import { Component, OnInit, ViewEncapsulation } from '@angular/core';
+import { MomBlogService } from '@momentum/frontend/data';
+import { finalize } from 'rxjs/operators';
+import * as txml from 'txml';
```
tsa96, line 4: From txml docs (https://www.npmjs.com/package/txml) under Installation:
>For specially small builds using modern module bundlers like rollup or webpack you can import txml/txml or > txml/dist/txml. This will not add the transformStream into the bundle and with that exclude the Node.js files.

apps/frontend/src/app/pages/dashboard/community/community-news/community-news.component.ts
tsa96: Nice, new approach is much cleaner

tsa96 (approved):
Okay looking good, I'll get this in once my dependency upgrade branch is in, should resolve build failing in CI.

tsa96 (changes requested):
Wait, Prettier is still failing though, run `nx format` then `fixup` any changes

package.json
```diff
@@ -70,6 +70,7 @@
     "simple-icons": "8.12.1",
     "steam-appticket": "1.0.1",
     "tslib": "2.5.0",
+    "txml": "^5.1.1",
```
Gocnak, line 73: Pin versions to specific numbers where possible.

tsa96 (approved):

PR #781, opened 2023-07-06
Refactor backend repo layer

### LINKED ISSUE #745: Refactor backend repo layer
Something from my notes I wanna get a card for since it'll be a while til I get to it.

Current Prisma layer is a load of shit, messes with return types and actually means we do fewer SELECTs which would save RAM.

Kind of waiting for Prisma Client Extensions to be released, but not completely sure if worth using, ran into some issues when I spent a day messing with them (tbh I don't remember the issue that well and lost the stash from when I was working on it...).

I might have another go with forwarding return types one last time and if that works, consider keeping the current repo module methods  but always using the exact same args as their respective Prisma functions. This would actually only be a few hours work, the repo code is very easy to change, just needs some care in going through and finding places we can use SELECTs more.

tsa96:
Removes all the gross db wrapper logic and uses Prisma directly. Uses the new Client Extensions feature to provide some custom methods, and set define some default user creation behaviour - we can continue to use that in the future for wherever we want to be extra careful that some DB call is always done a certain way (though typically we manage this fine by avoiding duplicate service logic).

I also cleaned up the DB package a bit so we can fit the client extensions into the `db` package - so if anything besides the `backend` app (maybe some future microservices, idk) can use the same extensions.

Closes #745


tsa96:
Something screwed up with CI, maybe also the db:generate script still. Going to work on this tomorrow

Gocnak (approved):
Nice stuff, way more straightforward

Issue #292, opened 2019-02-23
Allow user account deletion


Gocnak:
In accordance with GDPR, we must be able to comply with the right to be forgotten.

SteamIDs are the only automatically collected user-identification, other than the data explicitly set by the end user on the site. Allowing users to delete their account, and all related data with it.

tsa96:
Several possible approaches
Create new role `DELETED`, assign to that user, remove all their info, rename to "Deleted User"
Pros: 
- Minimal DB operations
- Easy to do
Cons:
- Potentially lots of deleted users sitting in DB
- Extra role to handling in some places

Alternative: use placeholders
Rather than new role, just change them to placeholder and remove everything
Cons:
- You lose uniqueness of placeholder aliases
  - Unless we merge users together, but that breaks constraint of one user having one ranked run per track per submode per map
- Placeholders now have runs on leaderboard??

Do we want to just allow runs also be deleted? That way we could actually delete users, or make them a placeholder if they have any maps submitted. Gives even more privacy but I can't imagine us having any kind of legal obligation. New ranking system could probably handle it fine though.

My preference is just towards top method (new `DELETED` role)

@Gocnak thoughts?

tsa96:
Just realised that admins can delete users already. Prisma schema is configured to `onDelete: SetNull` for _runs_, but `Cascade` for _ranks_. Which doesn't really make sense - we don't recalculate all the `rank` fields on the `Rank` table, we could have a leaderboard that's just missing a rank. This is likely to mess with pagination, really not something we want.

So how about we do the same approach (with `DELETED` role) as above, in `users.service.ts` and have both the `/user` and `/admin` endpoints use that same logic?

Gocnak:
Deleted role sounds good 👍

tsa96:
Okay great!

Here's a basic `BadgeDeleted.svg` for the frontend.
```svg
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="64" height="64">
  <defs>
    <linearGradient id="a">
      <stop offset="0" style="stop-color:#ffb446;stop-opacity:1"/>
      <stop offset="1" style="stop-color:#ffec3c;stop-opacity:1"/>
    </linearGradient>
[...]

PR #791, opened 2023-07-18
Allow user deletion

GordiNoki:
This PR closes issue https://github.com/momentum-mod/website/issues/292

### DIFF
--- a/apps/backend-e2e/src/admin.e2e-spec.ts
+++ b/apps/backend-e2e/src/admin.e2e-spec.ts
@@ -508,17 +508,184 @@ describe('Admin', () => {
 
       afterEach(() => prisma.user.deleteMany());
 
-      it('should delete a user', async () => {
+      afterAll(() => db.cleanup('map', 'rank', 'run', 'report', 'activity'));
+
+      it('should delete user data, leaving placeholder', async () => {
+        const followeeUser = await db.createUser();
+        const map = await db.createMap();
+        let userToBeDeleted = await prisma.user.create({
+          data: {
+            roles: Role.MAPPER,
+            bans: Ban.BIO,
+            steamID: 12345,
+            alias: 'User to be deleted',
+            avatar: 'yeeeee',
+            country: 'NA',
+            userAuth: { create: { refreshToken: 'yeeeee' } },
+            profile: {
+              create: { bio: 'yeeeee', socials: { RandomSocial: 'bemyguest' } }
+            },
+            userStats: { create: { totalJumps: 100 } },
+            submittedMaps: { connect: [{ id: map.id }] },
+            mapCredits: {
+              create: { type: MapCreditType.AUTHOR, mapID: map.id }
+            },
+            mapFavorites: { create: { map: { connect: { id: map.id } } } },
+            mapLibraryEntries: { create: { map: { connect: { id: map.id } } } },
+            mapReviews: {
+              create: {
+                map: { connect: { id: map.id } },
+                text: "That's a good map"
+              }
+            },
+            follows: {
+              create: { followed: { connect: { id: followeeUser.id } } }
+            },
+            followers: {
+              create: { followee: { connect: { id: followeeUser.id } } }
+            },
+            mapNotifies: {
+              create: { map: { connect: { id: map.id } }, notifyOn: 8 }
+            },
+            runSessions: {
+              create: {
+                trackNum: 1,
+                zoneNum: 1,
+                track: { connect: { id: map.mainTrackID } }
+              }
+            }
+          },
+          include: {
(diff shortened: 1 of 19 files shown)


apps/backend-e2e/src/admin.e2e-spec.ts
```diff
@@ -508,17 +508,194 @@ describe('Admin', () => {
...
-      it('should delete a user', async () => {
+      afterAll(
+        async () =>
+          await Promise.all([
+            prisma.map.deleteMany(),
+            prisma.rank.deleteMany(),
+            prisma.run.deleteMany(),
+            prisma.report.deleteMany(),
+            prisma.activity.deleteMany()
+          ])
```
tsa96, line 519: We have a method in `DbUtil` for handling this, using a Prisma `$transaction` to let Prisma handle the deletes and avoid gridlocks (doing these all in parallel is quite likely to cause that, especially for any cascade delete relations).

tsa96: Also, `async/awaits` are unnecessary, Jest will handle any promises you return to it.

apps/backend/src/app/modules/admin/admin.service.ts
```diff
@@ -19,10 +19,14 @@ import {
 import { Bitflags } from '@momentum/bitflags';
 import { Role } from '@momentum/constants';
 import { DbService } from '../database/db.service';
+import { UsersService } from '../users/users.service';
 
 @Injectable()
 export class AdminService {
-  constructor(private readonly db: DbService) {}
+  constructor(
```
tsa96, line 26: No need for the `deleteUser` method here, just call `delete` in the `UsersService` directly from the AdminController. Otherwise we end up with a spaghetti of service methods calling each other and nothing else.

apps/backend/src/app/modules/users/users.service.ts
```diff
@@ -281,6 +281,46 @@ export class UsersService {
...
+      profile: {
+        update: { bio: '', socials: {} }
+      },
+      mapFavorites: { deleteMany: {} },
+      mapLibraryEntries: { deleteMany: {} },
+      activities: { deleteMany: {} },
+      follows: { deleteMany: {} },
+      followers: { deleteMany: {} },
+      notifications: { deleteMany: {} },
+      runSessions: { deleteMany: {} }
```
tsa96, line 307: These shouldn't be needed - our Prisma schema should be configured Postgres to cascade delete all of these. If any aren't being deleted, add an `onDelete: Cascade` to the relation.

apps/frontend/src/app/components/delete-user-dialog/delete-user-dialog.component.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+//   beforeEach(() => {
+//     fixture = TestBed.createComponent(DeleteUserDialogComponent);
+//     component = fixture.componentInstance;
+//     fixture.detectChanges();
+//   });
+
+//   it('should create', () => {
+//     expect(component).toBeTruthy();
+//   });
+// });
```
tsa96, line 28: Why comment this out? I have a branch in review that gets frontend unit tests working, and this looks fine. If you're having any issues with `NbDialogRef`, override it with `[{ provide: NbDialogRef, useValue: {} }]`. See my `refactor/frontend-auth-and-unit-tests` branch for examples of how to unit test Nebular dialogs.

tsa96 (changes requested):
Sorry, should've reviewed this yesterday but got bogged down with map submission work.

apps/backend/src/app/modules/users/users.service.ts
GordiNoki: I know that they cascade delete, and I wanted to use it. The issue is that we don't delete the user

apps/frontend/src/app/components/delete-user-dialog/delete-user-dialog.component.spec.ts
GordiNoki: Spec was commented out on confirm dialog component, so when I copied it I leff it as is, just changed class names. I will write spec then

apps/backend/src/app/modules/users/users.service.ts
tsa96: OH, I'm sorry, I'm such an idiot.

tsa96 (changes requested):
Okay looks good - though please let reviewers resolve threads they brought up.

tsa96 (approved):
Okay looks good - though please let reviewers resolve threads they brought up.

PR #797, opened 2023-08-05
fix(back): make /users GET search by alias case-insensitive

### LINKED ISSUE #796: Make /users GET search by alias case-insensitive
This is really really easy, see https://www.prisma.io/docs/concepts/components/prisma-client/case-sensitivity#postgresql-provider
Then write an E2E test that it successfully matches a name with differing case.

nihlen:
Closes #796


tsa96 (approved):
Looks good, nice work!

Issue #743, opened 2023-05-22
Consolidate duplicate API endpoints


tsa96:
This is something we maybe should've done during the rewrite, but didn't, to preserve compatibility with the frontend and game, Currently we have several endpoints that all do essentially the same job, in particular fetching maps and runs. Generally, they end up calling the same methods in the backend service layer, just with variations of query DTOs.

So, let's simplify and remove all the redundant ones. To do this: 
1. Go through all the controllers/generated swagger docs and look for endpoints returning the same DTOs
2. Merge their query DTOs and service logic together
3. Delete the controller declarations for the redundant endpoints, where we generally try to the endpoint with the URL that corresponds to the DTO it's returning. For example, we seem to have several endpoints for returning runs, one on `api/runs/` and another on `api/maps/<mapID>/runs`. The former can take a `mapID` as a query param, so we should delete the latter
4. Update the frontend's data services to only use the consolidate endpoint
5. Give me a list of everything you changed. I'll check everything in game code and consolidate anything there (this requires engine license)

The one caveat is permissions. As far as I know the only one that matters is the `api/admin/maps`, which additionally allows admins/mods to filter by map queue status and priority. Given that we'll return those values to anyone anyway, I don't think it really matters, but idk it's probably closely related to map submission system which we'll be working on soon, so lets discuss that one in Discord once we've started on map submission work.

tsa96:
Note that this is blocked temporarily, until I get the Nx branch in. Just wanted to get it written up whilst fresh in my mind.

tsa96:
Activities is so stupid, there's the `/user/activities GET` user-specific version, which *does* have filtration to not send `REPORT_FILED` activities, but the general `/activities GET` *doesn't* have it. Joining these probably makes sense (though I'm really tired atm, give it more thought), include that logic in there.

```ts
if (type === ActivityType.REPORT_FILED)
      throw new UnauthorizedException(
        'Only moderators may search reports (use the reports endpoint!)'
      );
```
(haven't tried this)
```ts
    it('should 409 if the consumer searches for REPORT_FILED activities', () =>
      req.get({
        url: `users/${user.id}/activities`,
        query: { type: ActivityType.REPORT_FILED },
        status: 409,
        token: token
      }));
```

tsa96:
Admin user updates don't support country and socials atm but complicated merging the two together and I don't want a bunch of duplicate code so leaving for now. Would like to merge but in such a way that the service logic doesn't become insane complicated

PR #819, opened 2023-08-12
Merge runs/ GET, maps/:mapID/runs GET

benjl:
Attention: anything that used `maps/:mapID/runs/ GET` or `users/:userID/runs GET` has to use `runs/ GET` instead now  
(Issue #743)


tsa96 (approved):
Looks good, I really should've been more critical when porting and just skipped them stuff. The game never queries for runs *at all*.

Fixed CI on this branch because I'm lazy, and also remove some models you forgot.

PR #820, opened 2023-08-12
Map submission part 1

### LINKED ISSUE #764: Rename model Map to MMap
Using the word "Map" for referring to game maps is stupid, and often clashes with the native Javascript `Map` data structure. In some files we do an "import as" and rename, but we should be consistent and rename the entire model (though model/DTO keys will stay the same). MMap makes sense to me, very similar to Map (keep in mind we have tons of stuff like MapCredit, MapType, we don't want to rename all those as well, this is enough to make the difference clear and we'll have JSDoc comments.

### LINKED ISSUE #811: Separate fileName field on MMaps
Since maps will be able to support multiple modes in the future, we're going to store two variants of the map name
- the `name`: name of the map with no gamemode prefix or random suffixes, e.g.  "arcane"
- the `fileName`: the full name with gamemode prefixes (optional) and other miscellaneous stuff, e.g. "bhop_arcane"

We haven't entirely decided on UI for this, but this approach gives us a lot more freedom. It also means we can stop storing a `fileKey` in database, which always has redundant "maps/<name>.bsp". We'll always store BSPs (and VMFs in the future) relatively to the `fileName`, just remove the redundant data.

### LINKED ISSUE #813: Map submission schema changes and new models
This is a blocker to practically all other work. I'm working on it on `feat/map-submission-part-1`, and aim to get a PR open in the next couple of days.

tsa96:
Closes #813 
Closes #764 
Closes #811 

Implements the underlying schema and models required for map submission - once this is in, all the new endpoints on the Map Submission project are available to be worked on (so @Gocnak if you could review fairly quickly that'd be hugely appreciated, this probably the only PR that we need to get in ASAP).

Sorry if this is a bit hard to review, this is mostly refactors needed before we get to the brunt of new map submission stuff.

### DIFF
--- a/apps/backend-e2e/src/admin.e2e-spec.ts
+++ b/apps/backend-e2e/src/admin.e2e-spec.ts
@@ -508,7 +508,7 @@ describe('Admin', () => {
 
       afterEach(() => prisma.user.deleteMany());
 
-      afterAll(() => db.cleanup('map', 'rank', 'run', 'report', 'activity'));
+      afterAll(() => db.cleanup('mMap', 'rank', 'run', 'report', 'activity'));
 
       it('should delete user data. leaving user with Role.DELETED', async () => {
         const followeeUser = await db.createUser();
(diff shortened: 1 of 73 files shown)


Gocnak (approved):
Nice work!

PR #824, opened 2023-08-17
Map Credits refactor

### LINKED ISSUE #415: Update map credits types

### LINKED ISSUE #808: Support arrays of DTOs on MapCredit and MapImage create/update/delete endpoints
Currently when we update map images or credits on the frontend, we have to send a big batch of individual HTTP requests for each change. I really dislike this approach, as it spams our backend - with writes!

Array checks are one of least upsetting type-checks in JavaScript, we can easily allow these endpoints to receive and return `DTO | DTO[]` and do a type guard. The actual create/update/delete logic can be done in separate private functions which are either called once (individual DTOs) or multiple times (arrays).

We could *only* accept arrays, but we *are* an API, accepting single values seems reasonable.

Once the backend is refactored, update the frontend code to use arrays (singleton arrays ares fine of course). Should make that code significantly simpler, especially given some of the RxJS `forkJoin`s we have to do because of the current system.

Note, for the new `MapTestingRequest` endpoints we'll be supporting this pattern out of the box.

tsa96:
Making credits less insane on both front and backend.

On the backend, we had unnecessarily complicated endpoints and a pointless ID for MapCredits in the schema. This refactors out that ID, and replaces the multiple POST/PATCH/DELETE endpoints for individual MapCredits with a single PUT that takes an array of all the credits for a map. As you'll see, this *drastically* simplifies code on both front and backend.

On the frontend, we had unnecessarily complicated (there's a theme here) and sometimes duplicated logic for handling objects containing arrays of MapCredits (which is needed for drag-and-drop, took me a while to figure that out...). Also, we were using an extra layer of components for something that can be easily handled using an `*ngFor`.

With all that cleaned up, I was finally able to complete https://github.com/momentum-mod/website/issues/415. We'll need to update in game code and Panorama at some point - have made an issue here: https://github.com/momentum-mod/game/issues/2092

Closes #415 
Closes #808 - this handles the MapCredits side of it completely, and I don't think MapImages are as important, honestly unless we have an major issue for map images in the future (less likely, since we don't have to do all that weird filtration and stuff in frontend code), I think we can leave it.


tsa96:
Absolutely no idea why CI is failing here, nothing changed that should affect it. Maybe Github/NPM issue? Going to try again in a few hours.

apps/backend/src/app/modules/maps/map-credits.service.ts
```diff
@@ -100,72 +55,99 @@ export class MapCreditsService {
...
     });
+    if (!map) {
+      throw new NotFoundException('Map not found');
+    }
 
-    if (!credit || !credit.mmap || !credit.user)
-      throw new NotFoundException('Invalid map credit');
+    await this.checkCreditChangePermissions(map, loggedInUserID);
+
+    // TODO: Wrap in transaction!
```
Gocnak, line 80: Should this TODO be tackled in this PR? Or should there be an issue?

tsa96: In next PR

tsa96: Required some DB refactoring, couldn't do in this, is in next

Gocnak (approved):

Issue #757, opened 2023-06-13
Use Prisma interactive transaction API for sequential DB operations


tsa96:
I've not taken enough care to ensure that certain operations that perform multiple DB operations will fail gracefully. Currently, if something errors during the map creation service logic, the database can be left in a state with a partially added map. Really not good!

The main reason for this is because of how annoying Prisma can be with creating complex nested data - it doesn't currently support nested `createMany`s (and probably never will, devs seem to be deliberately ignoring the issue https://github.com/prisma/prisma/issues/5455). So we do this [nasty looping](https://github.com/momentum-mod/website/blob/822a1b375694f98da5553ed3ee996d67e3d1ec20/apps/backend/src/app/modules/maps/maps.service.ts#L214C5-L355) adding chunks of data for each couple of layers of nesting.

Fortunately Prisma now has this [interactive transaction API](https://www.prisma.io/docs/concepts/components/prisma-client/transactions#interactive-transactions) that lets us wrap all this in one big transaction. Shouldn't be a difficult refactor, just too busy to do right now.

tsa96:
Our current setup for Prisma/Nest client extensions breaks the usage of `$transaction`, we need to move to a better approach to providing the extended PrismaClient within Nest, [this solution](https://github.com/prisma/prisma/issues/18628#issuecomment-1623189625) looks like it should work.

Gordi's work on user deletion would be utilising `$transaction` but can't, that should also be included in this issue. Worth looking for other queries that should do this as well, run submission probably needs that as well.

PR #825, opened 2023-08-18
Refactor/prisma transactions

tsa96:
Wraps any sequential DB writes in Prisma's [interactive transactions](https://www.prisma.io/docs/concepts/components/prisma-client/transactions#interactive-transactions) so if one operation fails, everything gets rolled back. 

Closes https://github.com/momentum-mod/website/issues/757


apps/backend/src/app/modules/admin/admin.service.ts
```diff
@@ -61,99 +61,101 @@ export class AdminService {
       include: { follows: true, followers: true }
     });
 
-    if (!user) throw new BadRequestException('Merging user not found');
+    return await this.db.$transaction(async (tx) => {
+      if (!user) throw new BadRequestException('Merging user not found');
```
Gocnak, line 65: This can probably be outside the transaction?

Gocnak (changes requested):

Gocnak (approved):

Issue #739, opened 2023-05-10
Remove Angular frontpage content entirely 


tsa96:
We've long since decided that we're going to redo the frontpage (momentum-mod.org) site, without using any major frameworks. Exactly when we'll fit this in is somewhat unclear, depends when both I and traz and find the time. That's more of an internal scheduling thing, doesn't need tracking here.

One thing that *does* need tracking here is how we untangle the current site from the rest of the Angular setup. Since we want the current site for work for now, I'm leaving the code untouched, but once we can finally nuke it, here's some things we'll want to do. Note, this is based on the massive upcoming Nx restructure I'm working on.

First, we restructure the `apps/frontend/src/app/pages/` directory. We don't need the `dashboard/` subdir in there, pull all the children out into `pages/`. The `frontend` Angular app that lives in `apps/frontend` effectively *will* be the dashboard, and nothing else.Any other sites we put in this monorepo would be a separate app, if we did some other Angular app for some reason, anything the two Angular apps both needed can be moved to a shared library in `libs/`.

The `not-found` and `outgoing` pages will want cleaning up:
- `pages/not-found/main` can get nuked. As above this app is purely for the dashboard now.
- `pages/not-found/dashboard` moves to just `pages/not-found`.
 - `outgoing` is also getting removed (https://github.com/momentum-mod/website/issues/752)

Will probably add to this further as I continue cleaning up the frontend.

tsa96:
Should note we currently have this annoying issue with `NbOverlay` not working, *apparently* because the whole app isn't wrapped in `<nb-layout>` (see [here](https://github.com/akveo/nebular/issues/943), though that doesn't seem to actually fix it. So currently lazy-loaded module components get wrapped in extra `<nb-layout>`s which is super ugly.

Once we do this issue, is done, maybe worth giving it another go. I tried wrapping the root app template in `<nb-layout>` to no avail, but felt kinda hacky. During this issue, we probably want to do that *anyway*, so worth having another fiddle with the `<nb-layout>` stuff again

tsa96:
`not-found-main` is the only page currently using `splash.webp`, be sure to remove that file during this.

PR #831, opened 2023-08-20
Initial frontend styling

tsa96:
This is by no means a complete restyle, frankly it doesn't look great (still better than current stuff). I think it's worth getting this work in now though - it's probably going to be a couple of months until I can focus on styling and this gives other contributors something closer to what our dashboard will look like in the future.

Main issue is font-size scaling, which effectively rescales *everything* on the site, and I've much rather contributors design stuff using regular 16pt scaling, even if some parts of the existing UI might look a bit worse with this change. Better they work with the correct scaling now and get stuff looking good, than they do something for the 14pt scaling, and we have to go in and tweak it again once we do the full restyle.

#### Old
Momentum_Mod_-_Google_Chrome20,08,23,234249

#### New
Momentum_Mod_-_Google_Chrome20,08,23,234226

### DIFF
--- a/apps/frontend/src/app/pages/dashboard/dashboard.component.scss
+++ b/apps/frontend/src/app/pages/dashboard/dashboard.component.scss
@@ -3,6 +3,7 @@
 @import '@nebular/theme/styles/global/breakpoints';
 
 @include nb-install-component() {
+  // TODO: Unfuck this
   .socials {
     position: absolute;
     bottom: 0.25em;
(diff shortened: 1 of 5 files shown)


apps/frontend/src/app/pages/dashboard/dashboard.component.scss
```diff
@@ -26,24 +27,3 @@
...
-    position: fixed;
-    left: 0;
-    right: 0;
-    bottom: 0;
-    height: 100vh;
-    z-index: -1;
-    will-change: transform;
-    min-height: 100vh;
-    min-width: 100vw;
-    background: url('/assets/images/splash.webp');
```
Gocnak, line 42: Is this image still used somewhere? Should it be deleted?

Gocnak (changes requested):

apps/frontend/src/app/pages/dashboard/dashboard.component.scss
tsa96: `not-found-main` uses it on the frontpage. I'd rather alter the frontpage as little as possible until we're ready to completely delete it, I can make a note in https://github.com/momentum-mod/website/issues/739 to remove it then.

Gocnak: Perfect thanks

Gocnak (approved):

PR #860, opened 2023-10-29
0.10.0 Zoning Refactor

### LINKED ISSUE #753: Remove pointless createdAt and updatedAt entries, use SELECTs better
We should not be store this data when it doesn't need to be, not should we be `SELECT`ing it when we don't need to.

So, go through the schema and remove these (and any other redundant columns) where possible.
Also, there's bound to be stuff that we don't need to fetch on some calls. Remember, it's more memory efficient only SELECT the stuff you need.

Related to https://github.com/momentum-mod/website/issues/745, I'm going to assign myself and do both at the same time

### LINKED ISSUE #858: Remove Gamemode.UNKNOWN
This has been removed from game code, shouldn't be in here.

tsa96:
This adds all the core backend changes for 0.10, in particular drastically reworking map submissions and run submission. I've kept the frontend changes to a minimum, the leaderboards and map description stuff could be significantly improved. I'm going to jump over to frontend very soon.

Apologies for such a monster PR, since we have such strong types deriving from DB schema plus heavy testing, and 0.10 changes the schema so drastically, this was an enormous amount to work to get something that passes CI and can be merged into main.

Closes #753 
Closes #858


tsa96:
Couple of small things I need to do to fix CI - I'll get that finished tomorrow.

libs/formats/src/zone/zone-validator.ts
```diff
@@ -0,0 +1,325 @@
...
+export const MIN_COORD_FLOAT = -MAX_COORD_FLOAT;
+
+export const MAX_ZONE_REGION_POINTS = 64;
+
+export const MAX_TRACK_SEGMENTS = 256;
+export const MAX_SEGMENT_CHECKPOINTS = 256;
+export const MAX_ZONES_ALL_TRACKS = 512;
+
+export const MAX_STAGE_TRACKS = MAX_TRACK_SEGMENTS;
+export const MAX_BONUS_TRACKS = 256;
```
jason-e, line 24: I've changed these three `256` to `255` so I can fit the values in a single byte while 1-indexing (e.g. the first stage has `majorNum == 1`).

libs/formats/src/zone/zone-validator.ts
```diff
@@ -0,0 +1,325 @@
...
+          `Volume ${volumeIndex} region ${regionIndex} does not have enough points`
+        );
+
+      if (region.points.length > MAX_ZONE_REGION_POINTS)
+        throw new ZoneValidationError(
+          `Volume ${volumeIndex} region ${regionIndex} has too many points`
+        );
+
+      // Using non-strict as JS coerces to value === undefined || value === null
+      if (region.bottom == null || region.bottom >= FLT_MAX) {
```
jason-e, line 130: We don't *really* need this `FLT_MAX` check, since it's just used in C++ to detect that the JSON field was null or non-existent. If the field is actually present and set to `FLT_MAX`, it would be caught by the out-of-bounds check below, which would make more sense anyway.

libs/formats/src/zone/zone-validator.ts
```diff
@@ -0,0 +1,325 @@
...
+        );
+      }
+
+      if (region.height == null || region.height <= 0) {
+        throw new ZoneValidationError(
+          `Volume ${volumeIndex} region ${regionIndex} has no height`
+        );
+      }
+
+      if (region.height > MAX_COORD_FLOAT * 2) {
```
jason-e, line 150: I don't do this check in C++, though I guess it wouldn't hurt to add it to catch corrupted data (even if a massive height value would technically still work ok). And maybe it should go a step further and instead check `region.bottom + region.height > MAX_COORD_FLOAT` ?

libs/formats/src/zone/zone-validator.ts
```diff
@@ -0,0 +1,325 @@
...
+        start,
+        trackName,
+        `segment ${segmentIndex} start zone`,
+        volumes,
+        true
+      );
+
+      // TODO: because of the .shift this doesn't include start zone. Make
+      // sure C++ version is the same.
+      if (checkpoints.length > MAX_SEGMENT_CHECKPOINTS)
```
jason-e, line 307: C++ does in fact count the first element in `checkpoints` toward the limit, which matters because the limit keeps `minorNum` exactly within the limits of a byte (when 1-indexing). C++ handles the segment start in the same loop now:
```c++
FOR_EACH_VEC( track.zones.segments, segmentIndex )
{
	const TimerSegmentDef &segment = track.zones.segments[segmentIndex];

	if ( segment.checkpoints.Count() > MAX_SEGMENT_CHECKPOINTS )
		return TimerDefsError( "Track %s segment %i has too many checkpoints", trackName, segmentIndex );

	FOR_EACH_VEC( segment.checkpoints, checkpointIndex )
	{
		if ( !CheckZone( segment.checkpoints[checkpointIndex], trackName,
			CFmtStr( "segment %i timer zone %i", segmentIndex, checkpointIndex ),
			volumes, volumeIsUsed, checkpointIndex == 0 ) )
			return false;
	}
}```

File: libs/db/src/prisma/schema.prisma [hunk 5]
```diff
@@ -374,6 +379,174 @@ model MapReviewComment {
...
+  /// sideways leaderboard and unstyled leaderboard. This field is used to "tag"
+  /// runs one leaderboard as also qualifying for another, e.g. the sideways 
+  /// run would get a "Sideways" tag on the unstyled leaderboard.
+  // This is an array because (a) it allows easier queries for specific flag(s)
+  // in Prisma (writing the paginated PastRuns get query raw would be really
+  // rough, though we could if absolutely needed), (b) we could conceivable
+  // exceed 64 flags at some point. This is of course a very large table where
+  // we care about size a great deal, though we expect the *majority* of runs
+  // If we need to cut back on the size of this table in the future, we could
+  // submitted not to have any flags, so it's probably pretty okay.
```
jason-e, line 468: Looks like these comment lines got swapped

libs/db/src/prisma/schema.prisma
```diff
@@ -374,6 +379,174 @@ model MapReviewComment {
...
+
+  // Tempting to add a (userID, createdAt) index here for e.g. user profile
+  // page, but the PK index starting with `userID` will do some of the work
+  // anyway, proooobably not worth the space. Worth testing out on live DB in
+  // future!
+
+  // Unique index starting with userID then profile, for queries like a user's
+  // run history.
+  @@id([userID, gamemode, style, mapID, trackType, trackNum])
+  @@index([mapID, gamemode, trackType, trackNum, style, time(sort: Asc), createdAt(sort: Desc)])
```
jason-e, line 488: `createdAt` should be ascending (the comment on like 422 gets this correct)

libs/constants/src/maps/incompatible-gamemodes.map.ts
```diff
@@ -0,0 +1,12 @@
+import { Gamemode as G } from '../';
+
+export const IncompatibleGamemodes: ReadonlyMap<G, G[]> = new Map([
```
jason-e, line 3: Is this used anywhere yet? I'm not sure how to interpret this structure.

tsa96: Yes, this is used when generating leaderboards during map submission. The idea is to generate leaderboards for every mode the submitter suggests that not *certainly* incompatible. So surf comes up a lot since any non-surf map you could maybe play in surf should just be played in bhop instead.

tsa96: See `getCompatibleLeaderboards` in leaderboardHandler.util for usage

libs/formats/src/zone/zone-validator.ts
tsa96: Okay, check is now `checkpoints.length + 1 > MAX_SEGMENT_CHECKPOINTS`

libs/db/src/prisma/schema.prisma
tsa96: Or rather I just started deliriously rambling - update comment should make more sense.

jason-e: It makes perfect sense if they were swapped!

tsa96: I'm confused, can you write exactly what you mean?

To expand on the comment a bit, Prisma schemas don't supposed nullable arrays for some reason (it's a much requested), so I'm not sure what Prisma ends up entering into the table when using an array. Postgres will use a bit more data for empty arrays over NULL (https://dba.stackexchange.com/questions/223518/is-there-a-disk-space-impact-when-choosing-null-vs-empty-array-for-the-default) so there may be a small size benefit to a bitfield, even for runs with no flags (likely quite a major of runs). If we added pguint (considering it) this could be some array of 1-byte uints which is probably very space-efficient (Postgres arrays use `varlena` under the hood which is IIRC on disk is essentially just an integer for the number of items followed by the items themselves), maybe in some cases runs with a single 1-byte flag plus extra array stuff ends up less than the 8 bytes for the bitfield.  Plus we could *maybe* have >64 styles (most we could pack in a single int) but unlikely to ever be >256.

Sorry dunno how much sense that makes, half asleep so more delirious rambling I suppose

jason-e: Don't worry about it, it's not a big deal and I'm not seeing what I saw before anyway (hopefully I didn't imagine it).

As for the best way to store tags, it seems like all the options are similar size-wise in the grand scheme of things, so I would just go with whatever is easiest to work with. I doubt we will ever have very many styles anyway, especially since there are downsides to having tons of small leaderboards.

Also, I think all runs should be tagged with their style, including "normal" runs. Otherwise it might be tricky to filter for those runs, especially if other non-style tags are present. And eventually, "normal" will be a sub-style of segmented and TAS, so we can't necessarily omit the tag due to it being the "top level" style.

tsa96: Mhmm okay I suppose you're right - maybe programmatically to omit the normal style and infer whether or not it exists based on the Run has tags incompatible with it (eg if it's not TAS segmented but, can't we infer it's normal?). But def sounds annoying do account for across multiple codebases and it's a very small space optimisation that we could always make further down the line if we got really desperate about space.

libs/formats/src/zone/zone-util.ts
```diff
@@ -0,0 +1,7 @@
+import { MapZones } from '@momentum/constants';
+
+export const ZoneUtil = {
+  isLinearMainTrack: function (zoneData: MapZones): boolean {
+    return zoneData.tracks.main.zones.segments.length === 1;
```
Gocnak, line 13: Is this the right check? A linear track is a track with just minor checkpoints. This seems to be checking if it just has one singular segment? What if there's multiple minor checkpoints? Or, I suppose, what are "segments" in this context? Our doc where we drafted this only has tracks, zones, volumes, and regions.

Gocnak (commented):
I started looking but got to a point where I think terminology has changed from the documentation we have.

libs/formats/src/zone/zone-util.ts
tsa96: Ah, I moved this commit quite high as I wanted to get through the formatting lib changes first, but I suppose it means this code ends up before the types for MapZones is given. See constants/models/maps/MapZones.ts 

The current zone file spec uses "segments" and "checkpoints", for major and minor checkpoints respectively, I don't mind refactoring but we'd have to do here and in C++ ASAP. rio brought this up the terminology briefly in mom-dev red on 26/10

jason-e: @Gocnak The original doc wasn't necessarily totally precise yet and was generally still structured from a relational DB point of view. Now that we know this data is all JSON, our data can be more structured, and as we implemented things we realized grouping zones into "segments" each with a list of "checkpoints" fit the needs of the data and has multiple benefits. The segment number is the "major number" and the checkpoint number within the segment is the "minor number". The first "checkpoint" in a segment is a "major checkpoint" in the old terminology (aka the track start if it is the first segment). And yeah if the Main track is linear, it has just one segment.

Gocnak (approved):

libs/db/src/prisma/schema.prisma
```diff
@@ -374,6 +379,174 @@ model MapReviewComment {
...
+  mapID Int
+
+  gamemode  Int @db.SmallInt
+  trackType Int @db.SmallInt
+  trackNum  Int @db.SmallInt
+  style     Int @db.SmallInt
+
+  /// We don't store ticks in anticipation of future sub-tick system. If ticks 
+  /// are required, they can be calculated with `time * tickrate of gamemode`.
+  time Float @db.Real
```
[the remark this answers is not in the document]
tsa96: I'm going based off of rio's guidance, which was to use time. We certainly shouldn't store both time and ticks (or tickrate itself) given that they're just multiples of each other based on a constant value defined per-gamemode. I suppose I'm not sure I understand what you're suggesting, can you clarify?

libs/db/src/prisma/schema.prisma
```diff
@@ -374,6 +379,174 @@ model MapReviewComment {
...
+
+  createdAt DateTime @default(now())
+
+  pastRun   PastRun? @relation(fields: [pastRunID], references: [id])
+  pastRunID BigInt?  @unique
+
+  // Tempting to add a (userID, createdAt) index here for e.g. user profile
+  // page, but the PK index starting with `userID` will do some of the work
+  // anyway, proooobably not worth the space. Worth testing out on live DB in
+  // future!
```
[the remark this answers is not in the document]
tsa96: For this index specifically? I'm not sure how well that fits into a card since experimenting with indexes is something we should start doing fairly regularly once we've deployed and have a decently large dataset - we'll have analytics showing what queries are being used most often and what's taking the longest, so we should clone that DB from time to time and test performance with and without certain indexes. Exactly what index combinations we should test is a little hard to anticipate ahead of time so feel like a card would be a bit vague. We'll certainly not get the perfect combination of indexes the first time, we'll be tweaking this file frequently based on what seems to be performing well.

libs/db/src/prisma/schema.prisma
```diff
@@ -374,6 +379,174 @@ model MapReviewComment {
...
+  /// for multiple leaderboards, and an separate entry is made for each. 
+  /// For example, a sideways surf run (for some track) would be added to the
+  /// sideways leaderboard and unstyled leaderboard. This field is used to "tag"
+  /// runs one leaderboard as also qualifying for another, e.g. the sideways 
+  /// run would get a "Sideways" tag on the unstyled leaderboard.
+  // This is an array because (a) it allows easier queries for specific flag(s)
+  // in Prisma (writing the paginated PastRuns get query raw would be really
+  // rough, though we could if absolutely needed), (b) we could conceivable
+  // exceed 64 flags at some point. This is of course a very large table where
+  // we care about size a great deal, though we expect the *majority* of runs
```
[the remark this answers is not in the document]
tsa96: Good catch, updated it

PR #869, opened 2023-11-30
Frontend map submission form

tsa96:
This page isn't totally finished, but I've largely moved away from it in the last week as I've worked on broader frontend refactoring, and further map submission stuff. Didn't want to leave this sitting around, so let's get this in.

The info text can get finalized soon, ideally after a meeting where people can proof what I write. All the styling will be finalized once Nebular is finally gone.

Based on top of https://github.com/momentum-mod/website/pull/867, please review that first.

Closes https://github.com/momentum-mod/website/issues/603

PR #867, opened 2023-11-27
Misc backend refactors, dependency updates

tsa96:
This upgrades pretty much every dependency to latest (including Postgres).

Then a bunch of backend refactors that I've done recently, mostly whilst working on frontend.

PR #872, opened 2023-12-06
Constant value xp systems

tsa96:
As discussed, stores the XP system variables as hardcoded values in Typescript, moves over to a dedicated library that both front and backend can use, and removes the database table and admin endpoints used to store and modify the values.

On top of #867 and #869, those need reviewing first.


Gocnak (approved):
Prettier was not kind to the XP systems module file LOL

tsa96:
> Prettier was not kind to the XP systems module file LOL

This isn't Prettier, it's just that since all the XP systems stuff moved to a separate class literally all is needed for the backend is an `@Injectable` so DI can pick it up

PR #880, opened 2024-01-01
fix(back,db): sets the initial level for users to 0

### LINKED ISSUE #877: Refactor XP systems code to allow being level 0
The cosmetic XP calculations don't currently support a user being level 0. But being level 1 requires a specific amount of XP (21,000). Starting a user with 21,000 xp obviously makes no sense, and we don't want to significantly change the algorithm, so we should just tweak it to support starting users at level 0.

This requires two changes:
- update the algorithm to handle the level 0 case correctly
- update user creation level to start the user with level 0, xp 0, instead of its current values of 1 and 21,000

ianwillis98:
Closes #877 

image (2)


Gocnak:
Hey, just FYI you can update your branch and it'll show up in the PR automatically! No need to close and open new PRs.

PR #881, opened 2024-01-01
fix(back,db): sets the initial level for users to 0

### LINKED ISSUE #877: Refactor XP systems code to allow being level 0

ianwillis98:
Closes #877

### DIFF
--- a/apps/backend-e2e/src/prisma.e2e-spec.ts
+++ b/apps/backend-e2e/src/prisma.e2e-spec.ts
@@ -80,7 +80,7 @@ describe('Prisma Client Extensions', () => {
       expect(profile.userID).toBe(userID);
       expect(userStats).toMatchObject({
         userID: userID,
-        level: 1,
+        level: 0,
         cosXP: 0n,
         mapsCompleted: 0,
         runsSubmitted: 0,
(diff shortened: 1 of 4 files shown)


tsa96 (approved):
Looks good, CI is happy! And like Nick said on your last PR, rather than open a new PR, if you need to make a change and don't want to muddy the commit log, you can stage any other changes, then do `git commit --amend -no-edit` to amend those changes onto your original commit, then `git push --force` to overwrite your original commit on remote (i.e. Github). If you have multiple commits it's a little more complicated, you need to do a `git rebase --interactive` and do `fixup`s, but no need to worry about that here. If you're having any issues with Github, feel free to ask for help on here in your commit description (can you always amend/fixup) or in our Discord - Git isn't easy software to use and we're always happy to help.

I'm a little concerned by the fact the original starting XP was 20000, since I assume someone specifically chose that. But we're likely to tweak cosmetic XP a great deal in the future anyway, and it's not that major a change, let's just go ahead and get this in.

Congrats on your first PR!

tsa96:
Oh, and Nick if you're reading - I'm doing the eqiuv changes to the constants on red feat/mom-0.10 now.

PR #886, opened 2024-01-09
Fix seed script on latest Prisma

tsa96:
One of the recent Prisma updates seems to have changed how values passed to `undefined` behave, which as far as I can tell, only affects this script. We don't run this in CI at all, so it got through tests on Gordi's recent PR.

Gordi, this will probably fix any issue with work you're doing, so feel free to rebase this in as soon as you approve.

### DIFF
--- a/libs/db/src/scripts/seed.ts
+++ b/libs/db/src/scripts/seed.ts
@@ -482,24 +482,25 @@ prismaWrapper(async (prisma: PrismaClient) => {
       }
     });
 
-    await prisma.activity.createMany({
-      data: [
-        {
+    await prisma.activity.create({
+      data: {
+        userID: map.submitterID,
+        type: ActivityType.MAP_UPLOADED,
+        data: map.id,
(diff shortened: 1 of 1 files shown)


libs/db/src/scripts/seed.ts
```diff
@@ -482,24 +482,25 @@ prismaWrapper(async (prisma: PrismaClient) => {
...
+        data: map.id,
+        ...Random.createdUpdatedDates()
+      }
+    });
+
+    if (map.status === MapStatusNew.APPROVED) {
+      await prisma.activity.create({
+        data: {
           userID: map.submitterID,
           type: ActivityType.MAP_UPLOADED,
```
GordiNoki, line 498: This status should be MAP_APPROVED

GordiNoki (changes requested):

libs/db/src/scripts/seed.ts
tsa96: Woops, good spot!

PR #891, opened 2024-01-10
Check BSP files to be compressed with bspzip

### LINKED ISSUE #836: Require submitted BSPs be bspzipped
`bspzip` LZMA compresses each lump in a BSP, drastically decreasing filesize, and AFAIK has no runtime cost in-game. Hosting large files costs us money and it's not hard for mappers to `bspzip` their maps, so let's require them do it.

Parsing just the *header* of a BSP is relatively easy, I forget the exact data structure but there's entries in each lump section in the header with the compressed and uncompressed size (or something similar, SCell says it's doable so I trust him). Let's just parse that (we could pull the basic binary reader functions out of `@momentum/replay` if needed) and check compressed size is significantly lower.

Should be possible to integrate this in both frontend and backend. Adding a BSP file is on the first parts of the upload form, so do the check then and show an error if it's not zipped. Backend `/maps POST` should 400 if not zipped.

GordiNoki:
Closes #836

Frontend part was done by Tom in his big frontend PR. This one adds same checks to map upload endpoints on backend, but also includes some fixes for reading BSP header from blob logic,


tsa96 (approved):
Looks good!

PR #894, opened 2024-01-13
Disallow self follow

### LINKED ISSUE #892: You can follow yourself
Both the front and backend allow hitting the "follow" button on your own profile, and it really creates a follow from you to yourself.

I think this must have been something I did when refactoring the frontend, on the live frontend the button doesn't show up.

On the frontend, put a check on the profile page to not show the follow component if you're on your own profile. On the backend, put an explicit check in place (`followeeID !== followedID` or similar) and write an E2E test for it.

alexwaibel:
This change prevents a user from following themself by both removing the "Follow" button from the profile view for the logged in user and by checking that target and logged in user are not the same in the follow API.

Closes #892

I simply removed the follow button component when viewing the logged in user's profile as this is not a bug in production so there should not be a risk of real users having following themself. If desired I can rework this to remove only the "Follow" button leave the "Unfollow" button.

### DIFF
--- a/apps/backend-e2e/src/user.e2e-spec.ts
+++ b/apps/backend-e2e/src/user.e2e-spec.ts
@@ -717,6 +717,14 @@ describe('User', () => {
         });
       });
 
+      it('should 400 if the authenticated user is the target user', async () => {
+        await req.post({
+          url: `user/follow/${u1.id}`,
+          status: 400,
+          token: u1Token
+        });
+      });
+
       it('should 401 when no access token is provided', () =>
         req.unauthorizedTest('user/follow/1', 'post'));
     });
--- a/apps/frontend/src/app/pages/profile/profile.component.ts
+++ b/apps/frontend/src/app/pages/profile/profile.component.ts
@@ -96,7 +96,7 @@ export class ProfileComponent implements OnInit, OnDestroy {
               });
             }
           }
-          this.isLocal = false;
+          this.isLocal = true;
           this.localUserService.refreshLocalUser();
           return this.localUserService.localUserSubject;
         }),
--- a/apps/backend/src/app/modules/users/users.service.ts
+++ b/apps/backend/src/app/modules/users/users.service.ts
@@ -450,6 +450,9 @@ export class UsersService {
     if (!(await this.db.user.exists({ where: { id: targetUserID } })))
       throw new NotFoundException('Target user not found');
 
+    if (localUserID === targetUserID)
+      throw new BadRequestException('Target user cannot be logged in user');
+
     const isFollowing = await this.getFollower(localUserID, targetUserID);
     if (isFollowing)
       throw new BadRequestException(


apps/backend-e2e/src/user.e2e-spec.ts
```diff
@@ -760,6 +768,14 @@ describe('User', () => {
           token: u1Token
         }));
 
+      it('should 400 if the authenticated user is the target user', async () => {
+        await req.patch({
+          url: `user/follow/${u1.id}`,
+          status: 400,
+          token: u1Token
+        });
+      });
```
GordiNoki, line 777: This test will fail, because PATCH method uses updateFollow, which wasn't modified. Also, this method is not protected at all, it easily throws 500 if relation doesn't exist, we should probably change that too, but I'll need to speak with Tom about it. For now, just remove this test.

apps/frontend/src/app/pages/profile/profile-follow/profile-follow.component.ts
```diff
@@ -21,6 +21,7 @@ export class ProfileFollowComponent implements OnInit {
   localFollowStatus: Follow; // The follow object of the local user following target user
   targetFollowStatus: Follow; // The follow object of the target user following local user
   checked = false;
+  isLoggedInUser: boolean;
```
GordiNoki, line 24: Creating new variable and checking if user is local is a way there, but actually parent component - profile.component already has that. And it even has check for local user for exactly this component on line 65 in profile.component.html. So the issue there is that the variable isLocal in profile.component doesn't work, so I would suggest to fix it there, rather than creating another variable.

GordiNoki (changes requested):

apps/backend-e2e/src/user.e2e-spec.ts
GordiNoki: Actually this test only works because body was not provided, which does return 400 error

apps/frontend/src/app/pages/profile/profile-follow/profile-follow.component.ts
```diff
@@ -31,6 +32,7 @@ export class ProfileFollowComponent implements OnInit {
   ngOnInit() {
     this.userSubject.subscribe((usr) => {
       this.user = usr;
+      this.isLoggedInUser = this.user.id === this.localUserService.localUser.id;
```
tsa96, line 35: Like Gordi says profile.component has this, just make a new `@Input' in this class and pass it in from the parent via the binding.

apps/frontend/src/app/pages/profile/profile-follow/profile-follow.component.ts
```diff
@@ -31,6 +32,7 @@ export class ProfileFollowComponent implements OnInit {
...
       this.localUserService
         .checkFollowStatus(this.user)
```
tsa96, line 37: There's no point calling `checkFollowStatus` if logged in, just early return.

tsa96 (changes requested):
Mostly looks good.

Sorry that this component is so messy, we shouldn't really be handling the `userSubject` subscriptions at all, it'd be betterif we just had a `User` input and used an `ngOnChanges` - I'll get an issue made to refactor all components like this in the future.

apps/backend-e2e/src/user.e2e-spec.ts
alexwaibel: Good catch, would it be worth asserting on error message in the POST test? I'm not right now because I was just following the format of some of the other test cases but it would ensure we're getting the expected 400 response, not just any 400.

apps/frontend/src/app/pages/profile/profile-follow/profile-follow.component.ts
alexwaibel: Using Gordi's approach I think we can just handle showing/hiding this component at the parent component level. Seems we don't need to pass anything in any more.

apps/frontend/src/app/pages/profile/profile-follow/profile-follow.component.ts
alexwaibel: Good callout, not relevant any more using Gordi's approach though as I've reverted any new logic added to this component.

apps/frontend/src/app/pages/profile/profile-follow/profile-follow.component.ts
alexwaibel: Good catch, I didn't realize that's what `isLocal` was referring to. I've updated the PR to use this approach, I agree this logic makes more sense in the parent component.

GordiNoki (approved):

apps/frontend/src/app/pages/profile/profile.component.ts
```diff
@@ -96,7 +96,6 @@ export class ProfileComponent implements OnInit, OnDestroy {
               });
             }
           }
-          this.isLocal = false;
```
tsa96, line 99: `isLocal` should be getting set to `true`. The logic is a bit convoluted by the above route param checks are determining if the logged in user's page or not, and setting `isLocal` accordingly, there's just no `else` block due to early return.

tsa96 (changes requested):

apps/frontend/src/app/pages/profile/profile.component.ts
alexwaibel: Is `isLocal` not already initialized to `true` [here](https://github.com/momentum-mod/website/blob/c7a72027ad0e8f8f23ae2d04f61263c37bb5f083/apps/frontend/src/app/pages/profile/profile.component.ts#L68C3-L68C18)?

tsa96: It's initialised as true, but Angular will keep the instance of the component alive across multiple router changes, hence we're updating this in paramMap. If a user were to navigate to another user's profile, then back to their own, isLocal would be false when it should be true

alexwaibel: Ahh I see, my angular component lifecycle knowledge is a bit rusty. Fixed in latest revision

tsa96: Yeah, I find the lifecycle stuff confusing sometimes, I'm not exactly sure when Angular will or won't reuse a behaviour.

But assuming that it *will* reuse a component instance, and since router events fire on first load, router events should handle every case we care about and thus more reliable than lifecycle events.

tsa96 (approved):

PR #895, opened 2024-01-15
test(back): adds tests for user creation in users-service

### LINKED ISSUE #868: Unit test user creation
User creation is obviously a very important, and we don't thoroughly test anywhere. Originally when we were only using E2E testing this wasn't possible, as we call Steam's API, but unit tests can easily mock those calls.

The auth calls to Steam are already well-tested, but the logic in `findOrCreateFromWeb`, `findOrCreateFromGame` and `findOrCreateUser` (all in `users.service.ts`) is not. All three of those methods need decent testing.

Nice easy issue, ofc tests aren't particular sexy, but this is a good introduction to unit tests for anyone unfamiliar with them.

ianwillis98:
closes #868

### DIFF
--- a/apps/backend/src/app/modules/users/users.service.spec.ts
+++ b/apps/backend/src/app/modules/users/users.service.spec.ts
@@ -1,0 +1,315 @@
+import { Test, TestingModule } from '@nestjs/testing';
+import { UsersService } from './users.service';
+import { mockDeep } from 'jest-mock-extended';
+import {
+  PRISMA_MOCK_PROVIDER,
+  PrismaMock
+} from '../../../../test/prisma-mock.const';
+import { EXTENDED_PRISMA_SERVICE } from '../database/db.constants';
+import { BadRequestException } from '@nestjs/common/exceptions/bad-request.exception';
+import {
+  ForbiddenException,
+  InternalServerErrorException,
+  ServiceUnavailableException
+} from '@nestjs/common';
+import { User } from '@prisma/client';
+import { AuthenticatedUser } from '../auth/auth.interface';
+import { SteamUserSummaryData } from '../steam/steam.interface';
+import { UsersGetAllQueryDto } from '../../dto';
+
+describe('UserService', () => {
+  let usersService: UsersService;
+  let db: PrismaMock;
+
+  beforeAll(async () => {
+    const module: TestingModule = await Test.createTestingModule({
+      providers: [UsersService, PRISMA_MOCK_PROVIDER]
+    })
+      .useMocker(mockDeep)
+      .compile();
+    usersService = module.get(UsersService);
+    db = module.get(EXTENDED_PRISMA_SERVICE);
+  });
+
+  it('should be defined', () => {
+    expect(usersService).toBeDefined();
+    expect(db).toBeDefined();
+  });
+
+  describe('getAll', () => {
+    it('should throw an error when passing both steamID and steamIDs', async () => {
+      const a: UsersGetAllQueryDto = {
+        steamID: '123456789',
+        steamIDs: ['123456789', '999999999']
+      };
+      await expect(usersService.getAll(a)).rejects.toThrow(BadRequestException);
+    });
+  });
+
+  describe('findOrCreateFromGame', () => {
+    it("should throw an error when the steamid's don't match", async () => {
+      const steamUserSummaryData: SteamUserSummaryData = {
+        avatar: '',
+        avatarfull: '',
+        avatarhash: '',
+        avatarmedium: '',
+        communityvisibilitystate: 0,
+        lastlogoff: 0,
+        loccountrycode: '',
+        personaname: '',
+        personastate: 0,
+        personastateflags: 0,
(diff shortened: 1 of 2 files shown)


apps/backend/src/app/modules/users/users.service.spec.ts
```diff
@@ -0,0 +1,286 @@
...
+        steamid: '123456789'
+      };
+      jest
+        .spyOn(usersService['steamService'], 'getSteamUserSummaryData')
+        .mockImplementationOnce(() => Promise.resolve(steamUserSummaryData));
+      await expect(
+        usersService.findOrCreateFromGame(BigInt(999999999))
+      ).rejects.toThrow(BadRequestException);
+    });
+    it('should throw an error when getSteamUserSummaryData throws and error', async () => {
```
tsa96, line 76: Nit: these files can get dense as it is, really worth adding an extra newline between outermost function calls and similar like this.

apps/backend/src/app/modules/users/users.service.spec.ts
```diff
@@ -0,0 +1,286 @@
...
+        realname: '',
+        timecreated: 0,
+        steamid: '123456789'
+      };
+      jest
+        .spyOn(usersService['steamService'], 'getSteamUserSummaryData')
+        .mockImplementationOnce(() => Promise.resolve(steamUserSummaryData));
+      const spy = jest.spyOn(usersService, 'findOrCreateUser');
+      await usersService.findOrCreateFromGame(BigInt(123456789));
+      expect(spy).toHaveBeenCalled();
```
tsa96, line 108: Here and in some other instances below you can use `toHaveBeenCalledWith` and pass in a specific object, should check it matches. One of the major parts of these functions is creating the appropriate Prisma object so worth testing that.

apps/backend/src/app/modules/users/users.service.spec.ts
```diff
@@ -0,0 +1,286 @@
...
+    });
+    it('should throw error when using a limited account if the setting is on', async () => {
+      const nonExistingUser: User = undefined;
+      db.user.findUnique.mockResolvedValueOnce(nonExistingUser as any);
+
+      jest
+        .spyOn(usersService['config'], 'getOrThrow')
+        .mockImplementationOnce(() => {
+          return true;
+        });
```
tsa96, line 268: I wouldn't usually bring this up but since I think you said you were new to JS, I just want to point out that this is equivalent to `.mockImplementationOnce(() => true )` - if an arrow function is a single expression (no block/semicolons) it just returns the result of the expression,

tsa96: Oh actually, this can just be `mockReturnValueOnce(true)`, as you're doing in some places above!

apps/backend/src/app/modules/users/users.service.spec.ts
```diff
@@ -0,0 +1,286 @@
...
+      jest
+        .spyOn(usersService['steamService'], 'isAccountLimited')
+        .mockImplementationOnce(() => {
```
tsa96, line 271: `mockResolvedValueOnce`. You can think of resolving as the async (so Promises, "pull") version of returning (so functions, "push").

tsa96 (changes requested):

apps/backend/src/app/modules/users/users.service.spec.ts
ianwillis98: Ah, thanks for pointing that out! `mockReturnValueOnce` is definitely nicer in these situations.

apps/backend/src/app/modules/users/users.service.spec.ts
ianwillis98: I was wonder what the difference between `mockReturnedValue` and `mockResolvedValue` was. Thanks for pointing it out!

apps/backend/src/app/modules/users/users.service.spec.ts
ianwillis98: That's a good point. I originally decided against using `toHaveBeenCalledWith` because the method calls are type checked, but the type checking won't ensure we are using the right values within the type. I will go ahead and add those.

ianwillis98:
It took me longer than I'd like to admit to get the commit history looking clean the checks to pass, but I think this PR should be in better shape now.

apps/backend/src/app/modules/users/users.service.spec.ts
tsa96: Yeah, also remember TypeScript types are just compile-time, and we can even just disable them when getting in the way within tests by casting to `any`. `toHaveBeenCalledWith` will be just check that the underlying JavaScript objects match up, which will use strict type checks but *JavaScript* strict type checks so like, making sure a string is a string, number is a number, nothing more complex than that.

tsa96:
Feel free to ping me in Discord with any questions relating to Git stuff, we're pretty strict with Git usage and it's not an easy program!

tsa96 (approved):

Issue #893, opened 2024-01-11
Clean up profile component


tsa96:
I left this page basically unfinished when removing Nebular, and it looks pretty terrible. The "follow" button's styling is totally borked, and the lists in the following, followers and runs tabs have basically no styling. And the overall layout just looks terrible! Why are avatars so enormous??

I'm very open to redesigning this page entirely, the tab component feels strange given how much free space we have. So if somebody wants to have a crack at redoing the layout, feel free to have go. We could also ask traz to do a whole mockup for it. But, smoothing out existing stuff is most important, so would also be very happy if someone wants to just do that! Let me know, and we'll split up the issue into two if needed.

tsa96:
Also, I broke the flag icons!!

Importing the flag-icons.min.css file into styles.css screams out errors: `Two output files share the same path but have different contents: media\es-ct.svg`. Don't have time to look at right now!

PR #906, opened 2024-02-08
Map info page

### LINKED ISSUE #777: Broken lightbox on recent Ng-gallery version
Just upgraded to new NPM package for https://github.com/MurhafSousli/ngx-gallery (we were waaaaay behind).

Trying to open the lightbox now errors with `Error: [NbOverlayContainerAdapter]: Layout not found.`. Wrapping the *entire* app in `<nb-layout>` does fix it, so we can give that another go once we get rid of the main page (see https://github.com/momentum-mod/website/issues/739 for more on NbOverlay issues).

I don't want to do the triple-wrapped setup now (each time adds 4 layers of junk to the DOM), going to just leave broken until we get rid of the main page. With the rate traz is working on mock-ups currently, we may even have that for 0.10.0!

### LINKED ISSUE #829: Map review component
New frontend component that fits in with the current map info pages (or maybe we do a full-width version like leaderboards section?).

Show paginated list of reviews, allow review submission, comment threads, resolving.

tsa96:
Total redesign of main map page, including leaderboards, submission details, and reviews.

Have some map submission form changes in here as well that I couldn't separate out that easily.

Closes #829
Closes #777


tsa96:
Oops, I'll do a Prettier run and re-push once someone reviews!

apps/frontend/src/app/theme/styles.css
GordiNoki, line 1: I'm not quite sure what exactly, but something broke the `.top-right-toolbar` in profile component. It has position absolute and was realtive to m-card for whatever reason, but now it's relative to body, and I have no idea why. The solution would be to add `position: relative` to m-card, but it may add other issues.

GordiNoki (changes requested):

apps/frontend/src/app/theme/styles.css
tsa96: The `profile` component is still using old styling stuff. There's an issue to redo it [here](https://github.com/momentum-mod/website/issues/893), in the meantime we don't care if it breaks. Especially when it has garbage classes like this:
```css
.top-right-toolbar {
  right: 1.25rem;
  position: absolute;
  z-index: 500;
}
```

GordiNoki (approved):

PR #911, opened 2024-02-22
Add devcontainer

alexwaibel:
We already encourage users to use docker to spin up containers for minio and postgres so I figured it could be useful to extend these configs such that devs are able to spin up the entire development environment in docker. This is completely optional, but hopefully this can help newer devs onboard more quickly and avoid comparability issues (for example incorrect node versions).

This config can also be utilized as a basis for running the project in places like CodeSandbox and GitHub Codespaces, which can sometimes be useful for sharing local builds, quickly spinning up branches and PRs in an env for testing, or developing on hardware like tablets where you might only have access to a browser. Since we utilize multiple containers, some extra steps [found here](https://github.com/codespaces-contrib/codespaces-multi-dev-container) are needed.

Setup steps for the repo when running in a devcontainer on your local machine:
1. Ensure docker, git, and VS Code are installed
2. Install the VS Code Dev Containers extension
3. Clone the project
4. Open the project root directory in VS Code
5. Run **Dev Containers: Reopen in Container** from the Command Palette (`F1`) and select `Momentum Mod API`
6. VS Code with then spin up the containers, reload the window, and connect to the container running the backend.
7. Once you see the files appear in the file explorer, open `.env` and add your `STEAM_WEB_API_KEY` as described [here](https://github.com/momentum-mod/website/wiki/Setup#environment-variables)
8. Use the Command Palette (`F1`) and run **Rebuild Container**
9. Once you see the `Nest application successfully started` log in the output, open a new VS Code window using **File > New Window** and open the project root in that new window
10. Run **Dev Containers: Reopen in Container** from the Command Palette (`F1`) and select `Momentum Mod Website`
11. Wait until you see the log `Local:   http://localhost:4200/` and open that address in your browser and you should see the Momentum Mod website. You should now have everything you need to contribute!

Once this is merged into main, we can use the following link instead of steps 2-5: 
[https://vscode.dev/redirect?url=vscode://ms-vscode-remote.remote-containers/cloneInVolume?url=https://github.com/momentum-mod/website](https://vscode.dev/redirect?url=vscode://ms-vscode-remote.remote-containers/cloneInVolume?url=https://github.com/momentum-mod/website) This automatically installs the Dev Containers extension, clones the repo, and opens the project in the container.

Other editors with Dev Container support can be used instead of VS Code can be used although I haven't tested these. See the list of supported editors [here](https://containers.dev/supporting#editors)

Troubleshooting:
- If for any reason you hit an error and the container exits, you can use the Command Palette (`F1`) and run **Rebuild Container** 
- If you see spinners on the website and Unauthorized errors in the API output, try hard refreshing the web page
- To stop the containers you can open the project root in a terminal on your host machine and run `docker compose down`

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,9 @@
   "license": "MIT",
   "scripts": {
     "postinstall": "nx run db:generate",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "serve:backend": "nx serve backend",
+    "serve:frontend": "nx serve frontend"
   },
   "dependencies": {
     "@angular/animations": "17.0.4",
--- a/.devcontainer/backend-container/devcontainer.json
+++ b/.devcontainer/backend-container/devcontainer.json
@@ -1,0 +1,28 @@
+{
+  "name": "Momentum Mod API",
+  "dockerComposeFile": [
+    "../../docker-compose.yml",
+    "../../docker-compose.override.yml",
+    "../docker-compose.devcontainer.yml"
+  ],
+  "service": "api",
+  "shutdownAction": "none",
+  "workspaceFolder": "/workspaces/${localWorkspaceFolderBasename}",
+  "features": {
+    "ghcr.io/devcontainers-contrib/features/nx-npm:1": {}
+  },
+  "customizations": {
+    "vscode": {
+      "extensions": [
+        "nrwl.angular-console",
+        "esbenp.prettier-vscode",
+        "firsttris.vscode-jest-runner",
+        "dbaeumer.vscode-eslint"
+      ]
+    }
+  },
+  "forwardPorts": [3000],
+  "initializeCommand": [".devcontainer/init"],
+  "postCreateCommand": "git config --global --add safe.directory ${containerWorkspaceFolder} && git config --global --add safe.directory ${containerWorkspaceFolder}/apps/frontend/src/app/theme/shared_styling && git submodule update --init",
+  "postStartCommand": "npm install && nx run db:push && nx run db:seed && npm run serve:backend"
+}
--- a/.devcontainer/docker-compose.devcontainer.yml
+++ b/.devcontainer/docker-compose.devcontainer.yml
@@ -1,0 +1,27 @@
+version: '3.8'
+
+services:
+  app:
+    image: mcr.microsoft.com/devcontainers/typescript-node
+    volumes:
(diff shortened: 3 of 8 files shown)


package.json
```diff
@@ -5,7 +5,9 @@
   "license": "MIT",
   "scripts": {
     "postinstall": "nx run db:generate",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "serve:backend": "nx serve backend",
```
alexwaibel, line 9: The devcontainer npx feature was not reliable here for some reason so I opted to have the container use these npm scripts as a wrapper.

tsa96:
Yeah providing a local container-only approach is reasonable, but wouldn't it be easier just dockerizing that entire process, rather than using devcontainers? I don't really know anything about devcontainers whatsoever, but keeping the number of dev technologies to a minimum is preferable, so I'd rather only offer devcontainers as an option if it can *replace* an existing requirement (e.g. Docker)

alexwaibel:
At their core, devcontainers are just a small editor config on top of the existing docker-compose stack that provides some metadata to your editor about which container in the compose stack to attach to, what plugins to install to the editor, what ports to forward, etc.

I opted to forward ports for example to avoid extra changes in this PR. Forwarded ports use an editor feature to make traffic appear to the container as if it came from localhost. We could instead modify the API process to listen on `0.0.0.0` instead of `localhost` and add `ports: 3000:3000` to the api service in docker-compose. Same with some of the commands in devcontainer.json for installing node modules and starting the service, we could instead author a Dockerfile that includes these directly and that container. I do think that could be worth exploring, maybe in a follow-up PR, but for now with this PR I've done my best to not modify the way the frontend and backend services behave at all since I'm new to the project and didn't want to disrupt any existing workflows.

tsa96:
Okay, that makes a lot of sense. And sorry, I've been so busy with other stuff lately I didn't sit down and look at devcontainers at all until just now. Glad that it's an open spec, and like you say they're just configurations for editors to follow, which is certainly helpful as a way to avoid getting hands dirty w Docker directly, having to choose what containers to attach to via CLI etc.

tsa96:
Heads up that we don't accept merge commits, please rebase!

tsa96:
Gonna close this since very stale and haven't had requests for this from anyone else. Appreciate the effort though!

Issue #878, opened 2023-12-31
Support viewing the dashboard without a Steam login


tsa96:
We get requested to add this basically every time the dashboard gets brought up. Now that the repo is in a much more organised state overall, and we have a much clearer authentication workflow, this should be pretty easy.

This isn't a complete spec, I'll add more in the future, just want to get something simple tracked now so I can reference in other places.

# Frontend
- Refactor `LocalUserService` and `AuthService` so it cleanly handles not having an active JWT or user account in the `localUserSubject` field.
- Replace the player card in the header component with a "Sign in with Steam" button (make sure to use the official one that Steam supplies).
- Create a `LoggedInGuard` guard that can protect any routes that don't make sense to view if you don't have a login - I can provide a full list of these if needed.
- Add an error handler in the HTTP wrapper I wrote that handles `429 - Too many requests` by sending a toast to users saying something like "You are being rate limited! Please sign in to increase your request capacity.".
- Disable some parts of forms like leaderboard filters when you're not logged in, with a similar warning as above. Generally we want to disable stuff that creates more expensive DB queries, such as `ORDER BY`s that don't have much indexing to help them.

# Backend
- Refactor auth service and all uses of the `@LoggedInUser` decorator to handle cases where the JWT is missing - would sure be nice to have `enableStrictNullChecks` on by this point...
- Disable the corresponding expensive filters if LIU is missing.
- Use Nest's Throttle module, and make it much more restrictive for requests that don't have a JWT - I assume it can track IP addresses as well? I haven't looked into it in a while, not sure what we'll use for caching - may have some general purpose Redis instances by then
- Remove the JwtAuthGuard from all the endpoints it makes sense to make available to users without a JWT.

If anyone wants to pick this up, definitely get in touch with me on Discord. This is a big change and I'm only providing loose details here, we'll want to write up something more precise in the future, probably after 0.10.0's release.

tsa96:
|Endpoint|Public|
|----|----|
|auth/web|FALSE|
|auth/web/return|FALSE|
|auth/game|FALSE|
|auth/refresh|FALSE|
|auth/revoke|FALSE|
|users|TRUE|
|users/{userID}|TRUE|
|users/{userID}/profile|TRUE|
|users/{userID}/activities|TRUE|
|users/{userID}/followers|TRUE|
|users/{userID}/follows|TRUE|
|users/{userID}/credits|TRUE|
|runs|TRUE|
|runs/{id}|TRUE|
|maps|TRUE|
[...]

PR #927, opened 2024-03-28
Support viewing the dashboard without a Steam login

GordiNoki:
Closes https://github.com/momentum-mod/website/issues/878


apps/frontend/src/app/components/map-list/map-list-item.component.ts
```diff
@@ -48,10 +48,18 @@ export class MapListItemComponent implements OnChanges {
...
   tierStyle: string;
 
+  protected isLoggedIn: boolean;
+
   constructor(
     private readonly localUserService: LocalUserService,
     private readonly messageService: MessageService
-  ) {}
+  ) {
+    this.isLoggedIn = this.localUserService.isLoggedIn();
```
tsa96, line 57: Rather than handling subscription yourself here, you can pipe the localusersubject changes, map to whether the user is logged in, then rather than call subscribe explicitly, use an async pipe.

apps/frontend/src/app/components/layout/header.component.ts
```diff
@@ -49,7 +49,7 @@ export class HeaderComponent implements OnInit {
   protected unreadNotificationCount = 0;
 
   constructor(
-    private readonly localUserService: LocalUserService,
+    protected readonly localUserService: LocalUserService,
```
tsa96, line 52: I prefer a method in this class that calls this, rather than exposing a service to a template

tsa96 (changes requested):
Read half of this, will do rest later this morning

apps/frontend/src/app/services/data/local-user.service.ts
```diff
@@ -62,17 +64,20 @@ export class LocalUserService {
     return this.authService.isAuthenticated();
   }
```
tsa96, line 66: My suggestion is to create a `BehaviourSubject<boolean>` here that emits whenever the user logs in or out. There's several places in your changes that I think this would be useful, see other comments

apps/frontend/src/app/components/map-list/map-list-item.component.ts
tsa96: To clarify, use the LocalUserService.LoginChanged BehaviourSubject<bool>, then expose the service. Then, in the template, you can do e.g. `@if (luService.LoginChange | async)`. I don't use async pipes enough, but since we have so much state attached to a single service, it's a very good case for them.

apps/frontend/src/app/pages/maps/browsers/map-browser.component.html
```diff
@@ -38,26 +38,28 @@
...
-      <m-n-state-button
-        class="mr-4 h-8 w-8 p-0"
-        [states]="[
-          { icon: 'flag-outline', color: null },
-          { icon: 'flag-outline', color: 'blue' },
-          { icon: 'flag-outline', color: 'red' }
-        ]"
-        formControlName="pb"
-      />
+      @if (isLoggedIn) {
```
tsa96, line 41: Same thing, async pipe.

apps/frontend/src/app/services/data/auth.service.ts
```diff
@@ -22,10 +22,8 @@ export class AuthService {
     localStorage.removeItem('accessToken');
     localStorage.removeItem('refreshToken');
     localStorage.removeItem('user');
-    // Redirects to frontpage for now, once we remove that from this project,
-    // we've have to do an ugly `window.location.href` redirect, at least until
-    // we have sections of the dashboard than can be used without a login.
-    this.router.navigateByUrl('/');
+    // Redirects to maps browser page, since it's the main page used by unauthroized user
```
tsa96, line 25: "unauthorized". Also worth noting this is temporary.

apps/frontend/src/app/pages/maps/map-info/map-info.component.html
```diff
@@ -54,24 +54,26 @@
...
-          class="btn btn-orange"
-          type="button"
-          (click)="editNotificationSettings()"
-          mTooltip="{{ notifications ? 'Disable Notifications' : 'Enable Notifications' }}"
-        >
-          <m-icon [icon]="notifications ? 'bell-alert' : 'bell-outline'" />
-        </button>
-        @if (!isSubmitter && !isModerator) {
-          <m-report-button class="text-lg" mTooltip="Report map" [reportType]="ReportType.MAP_REPORT" [reportData]="map.id.toString()" />
+        @if (isLoggedIn) {
```
tsa96, line 57: async pipe

apps/frontend/src/app/pages/home/home.component.ts
```diff
@@ -14,7 +15,13 @@ import { LocalUserService } from '../../services/data/local-user.service';
 export class HomeComponent {
   user: User;
 
-  constructor(private readonly localUserService: LocalUserService) {
+  constructor(
+    private readonly localUserService: LocalUserService,
+    private readonly router: Router
+  ) {
+    if (!this.localUserService.isLoggedIn()) {
```
tsa96, line 22: You're calling a regular just once from the constructor here. The only reason it works is because the login checks are syncronous and so happen before this constructor is called.

Better to subscribe to the loginchanged behavioursubject, and do the redirect it emits with `false`. That way if auth ever becames asyncronous, or you do a logout whilst on this page, it'll redirect.

apps/frontend/src/app/components/player-card/player-card.component.ts
```diff
@@ -16,7 +16,7 @@ import {
   styleUrl: './player-card.component.css'
 })
 export class PlayerCardComponent implements OnInit {
-  protected user: FullUser;
+  protected user?: FullUser;
```
tsa96, line 19: This is derived from I luservice's localUserSubject, which is a BS of `FullUser | null`. However, we know that this component should never be shown unless we have a LIU, due to the `@if {}`s in the template. If we don't, Angular should destroy the component. 

I think the correct approach is to keep the non-null type `user: FullUser`, and early return from the below `.subscribe` block if `user` is null there. I was considering throwing an error, since this component should be destroyed at roughly the same time, but I'm not sure exactly what order.

tsa96 (changes requested):
Gonna make some of these changes myself, as discussed in Discord

tsa96 (approved):
Okay, I've done the stuff I brought up myself, going to do a separate PR. Your backend stuff was spot on!

PR #936, opened 2024-04-20
Fix recent changes with docker and env variables

GordiNoki:
The solution for cookies is not great, but works for dev and should probably work for prod

### DIFF
--- a/apps/backend/src/app/modules/auth/auth.controller.ts
+++ b/apps/backend/src/app/modules/auth/auth.controller.ts
@@ -49,7 +49,7 @@ export class AuthController {
     private readonly steamOpenID: SteamOpenIDService
   ) {
     this.cookieOptions = {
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -5,9 +5,13 @@ services:
     container_name: postgres
     image: postgres:15-alpine
     restart: unless-stopped
+    env_file:
+      - .env
   minio:
     image: minio/minio
     container_name: minio
+    env_file:
+      - .env
```
tsa96, line 14: This should be in the override!

tsa96 (changes requested):

docker-compose.yml
GordiNoki: Done

tsa96:
Closing in favour of #937

PR #937, opened 2024-04-21
Fix production prep issues

tsa96:
Supercedes #936
- Fixed .env file not being loaded by dev docker environment (thanks Gordi)
- Changed how URLs/domains are handled
- Disabled ESLint commit hook because it takes an absolutely absurd amount of time. CI checks and encouraging people to use the editor plugin is enough IMO.
- Docker Image publishing *hopefully* works, followed your recommendation Nick.
- Random faff

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -3,6 +3,8 @@ version: '3.8'
 services:
   postgres:
     restart: unless-stopped
+    env_file:
+      - .env
     volumes:
       - ./data/postgres:/var/lib/postgresql/data
     expose:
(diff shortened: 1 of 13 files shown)


docker-compose.override.yml
GordiNoki, line 1: Probably mixed up my env_file entries and yours

GordiNoki (changes requested):

GordiNoki (approved):

Gocnak (approved):
It's probably not an issue yet but we may want to optimize the RUN steps in the docker file to minimize the size of the image. These changes are good so it's a future problem

PR #939, opened 2024-04-26
Multiple docker related fixes

GordiNoki:
This commit moves angular dependent code from shared libraries to frontend app, so backend docker container would be a bit smaller (~70 MB smaller). 
Also, I thought of moving prisma client generation to the same RUN entry as npm install, but we probably would update schema more often then packages. I wonder if we should move removing schema.prisma to the same RUN entry as client generation tho.


tsa96 (approved):
Yes I agree with putting the Prisma schema steps in one `RUN` command, can do that then merge yourself yourself

Gocnak:
If you're in the area of improving the dockerfile, please tackle #940 as well, one line change at the end of it after the `RUN`s.

PR #941, opened 2024-04-27
fix(meta): Run backend Dockerfile under non-root user

### LINKED ISSUE #940: Backend Docker image should run as non-root user
The backend Docker image we create https://github.com/momentum-mod/website/blob/main/Dockerfile should be run as the `backend` user created in the image. The `node` image also has the `node` user if we wanna just use that as well.

To do so, all we need is to add `USER node` or `USER backend` right before the `CMD` on the last line, like so:

```dockerfile
USER backend
CMD [ "node", "main.js" ]
```

Gocnak:
We also expose the port desired in the PORT env var via EXPOSE, so that our system running the backend can still talk to it

Closes #940

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,4 +16,6 @@ RUN npm --omit=dev -f install
 RUN node_modules/.bin/prisma generate --schema schema.prisma
 RUN rm schema.prisma
 
+EXPOSE 3000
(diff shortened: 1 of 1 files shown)


GordiNoki:
I took your changes to #939

PR #946, opened 2024-05-24
Add LOG_LEVEL env var

tsa96:
This is very helpful in tests, replacing the old "TEST_LOG_DEBUG" var. By default, E2E tests don't spew INFO logs anymore.


GordiNoki:
With these changes `nx serve backend` doesn't show INFO logs either, is it intended?

GordiNoki (approved):

PR #949, opened 2024-05-26
fix(back-e2e): help WebStorm load .env files

tsa96:
Osplo me and MinimumDonut have all been having issues with WebStorm loading environment variables lately. To be honest I'm not sure why it started happening. This just uses `dotenv` (which we have already) to load the file during E2E test setup.


GordiNoki (approved):
It's not only WebStorm that had problems with running tests, just running jest directly on file also had issues, but this does fix it.

PR #953, opened 2024-06-02
Add logic to prevent changing user alias on logout/login

### LINKED ISSUE #921: Backend resets user alias and country on each login
If you assign yourself a custom alias via the dashboard, sign out (browser discards JWT), sign in again (requests new JWT), we re-request your public Steam info, and update your profile with it, *even if the user already exists*.

Doing this for avatars makes sense, but if the account already exists, we shouldn't be setting the alias or country code.

This is like a 2 line fix + tests and I already see the offending code, doing myself ASAP.

kanoka42:
Date:      Fri May 31 10:58:39 2024 -0500

Implements fix for bug #921

Implements fix for bug #921 which resets the users alias on logou and back in. This fix
    - Implements a new regex to disallow purely whitespace based aliases
    - E2e testing on the new aliases
    - Logic to test if the user name is valid, or is the default value

Closes #921

### DIFF
--- a/apps/backend/src/app/modules/users/users.service.ts
+++ b/apps/backend/src/app/modules/users/users.service.ts
@@ -18,6 +18,7 @@ import {
   CombinedRoles,
   MapStatus,
   Role,
+  NON_WHITESPACE_REGEXP,
   UserMapLibraryGetExpand
 } from '@momentum/constants';
 import { Bitflags } from '@momentum/bitflags';
@@ -162,10 +163,12 @@ export class UsersService {
       where: { steamID: userData.steamID }
     });
 
+    const isValidAlias = user ? NON_WHITESPACE_REGEXP.test(user?.alias) : false;
+
     const input: Prisma.UserUpdateInput | Prisma.UserCreateInput = {
-      alias: userData.alias,
+      alias: isValidAlias ? user.alias : userData.alias,
       avatar: userData.avatar.replace('_full.jpg', ''),
-      country: userData.country
+      country: user?.country || userData.country
     };
 
     if (user) {
@@ -205,6 +208,9 @@ export class UsersService {
 
     // Strict check - we want to handle if alias is empty string
     if (update.alias !== undefined) {
+      if (!NON_WHITESPACE_REGEXP.test(update.alias)) {
+        throw new BadRequestException('Invalid alias');
+      }
       if (Bitflags.has(user.bans, Ban.ALIAS) && !asAdmin) {
         throw new ForbiddenException(
           'User is banned from updating their alias'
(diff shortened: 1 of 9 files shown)


kanoka42:
Closes #921

libs/constants/src/regexp/user-alias.regexp.ts
```diff
@@ -0,0 +1 @@
+export const USER_ALIAS_REGEXP = /^(?![\s028u{}]*$).+/;
```
GordiNoki, line 1: That's pretty weird regex. For example, nickname "u2u822" would be discarded, even though it is pretty random, it's not invalid

apps/backend/src/app/modules/users/users.service.ts
```diff
@@ -162,10 +163,16 @@ export class UsersService {
       where: { steamID: userData.steamID }
     });
 
-    const input: Prisma.UserUpdateInput | Prisma.UserCreateInput = {
-      alias: userData.alias,
+    const isValidAlias = USER_ALIAS_REGEXP.test(user?.alias);
+
+    const findOrCreateUserData = {
```
GordiNoki, line 168: Why did you put that in a separate variable?

GordiNoki (changes requested):
The commit name doesn't follow our naming conventions, make sure you have commit hooks working (run `npm install`). Also it worth doing test for the issue itself mentioned in #921, something like "should not update alias when user exists" in auth.e2e-spec.ts

apps/backend/src/app/modules/users/users.service.ts
kanoka42: Just for read-ability, but I can move it inside the object if you want

tsa96 (approved):
Reviewed together in voice

tsa96 (approved):

tsa96:
> The commit name doesn't follow our naming conventions, make sure you have commit hooks working (run `npm install`). Also it worth doing test for the issue itself mentioned in #921, something like "should not update alias when user exists" in auth.e2e-spec.ts

Some existing E2E tests were modified to check this, I think coverage is okay. Have asked for a unit test as well, then I think we're good.

kanoka42:
> The commit name doesn't follow our naming conventions, make sure you have commit hooks working (run `npm install`). Also it worth doing test for the issue itself mentioned in #921, something like "should not update alias when user exists" in auth.e2e-spec.ts

The tests had to be reworked to account for this specific change, so a few of the tests check for this issue now.

tsa96:
Woops!

libs/constants/src/regexp/user-alias.regexp.ts
tsa96: Just to follow up on this - this was a regex I wrote, based on a SO answer (which I tried to dig up just now but failed). The intention is to match any string that's not *entirely* whitespace. I recommend regexr.com for if you're struggling to understand a regex, see [regexr.com/816p3](this one) for a breakdown of what this one is doing.

PR #954, opened 2024-06-06
Feat/killswitches

### LINKED ISSUE #943: Killswitches For Core Functionality
A great admin tool we should have by 1.0.0 is killswitches; being able to "toggle off" core functionality of the backend.

E.g. a game-breaking bug is discovered and we need to shut off run submissions for a bit while we get a patch out.

Ideally the endpoint that would be tied to this killswitch would then return a 503 so it can be handled by the clients.

Kill switches I can think of:
- Run submission
- Map submission
- Authenticating with the backend in the first place

This list can be expanded upon with sub-level switches, e.g. turning off comments on map submission vs turning off map submission entirely.

In the future we could make the switches have 3 modes: ON, OFF, OFF for public, where just admins can still do things as to test if fixes are in place.

kanoka42:
This PR addresses issue #943.

### What this PR does
  - Creates a new module, service, controller, and guard. 
  - Implements the guard in certain routes to programmatically switch off or on certain endpoints.
  - Adds an endpoint for admins only that lets them update the current killswitches as well as retrieve which killswitches are currently active.
  - Implements the logic in certain deep functions for finer control of killswitch targeting (new user signups)
  - Updates unit tests to ensure the functionality of the killswitches

Closes #943


kanoka42:
Closes #943

apps/backend/src/app/modules/map-review/map-review.controller.ts
```diff
@@ -141,6 +151,10 @@ export class MapReviewController {
   // Not bothered with single comment GET, can't imagine every needing it.
 
   @Post('/:reviewID/comments')
+  @UseGuards(KillswitchGuard)
+  @Killswitch(KillswitchType.MAP_REVIEWS)
+  @UseGuards(KillswitchGuard)
+  @Killswitch(KillswitchType.MAP_REVIEWS)
```
tsa96, line 157: Duplicated lines

apps/backend/src/app/modules/killswitch/killswitch.service.ts
```diff
@@ -0,0 +1,83 @@
...
+      this.switches = {} as Killswitches;
+      await this.updateKillswitches(
+        Object.fromEntries(
+          Enum.values(KillswitchType).map((type: KillswitchType) => [
+            type,
+            false
+          ])
+        ) as Killswitches
+      );
+    }
```
tsa96, line 37: Maybe worth adding some logging in here. If the entry doesn't exist, let's log that we created it. If it *does* exist, and certain killswitches are enabled, let's output a log message (info level) for all the enabled switches.

apps/backend/src/app/modules/killswitch/killswitch.service.ts
```diff
@@ -0,0 +1,83 @@
...
+  private switches: Killswitches;
+
+  public async onModuleInit() {
+    const dbResponse = await this.db.config.findUnique({
+      where: { id: KILLSWITCH_CONFIG_ID }
+    });
+
+    const storedSwitches = dbResponse?.value as Killswitches;
+    if (storedSwitches) {
+      this.validateKillswitches(storedSwitches);
```
tsa96, line 25: Oh, this returns bool but we're not using it. If `validateKillswitches` returns false, let's log an **error** like "Stored killswitches failed validation" and init everything as false like the below code.

Sorry, I know I wrote this, so sort of reviewing my own stuff here!

apps/backend/src/app/modules/killswitch/killswitch.enum.ts
```diff
@@ -0,0 +1,7 @@
+// Enum is string based to make using as object keys easier
+export enum KillswitchType {
+  NEW_SIGNUPS = 'NEW_SIGNUPS', // This is more complicated to do
```
tsa96, line 3: This was a wip comment I left for you, can delete!

apps/backend/src/app/modules/killswitch/killswitch.service.ts
```diff
@@ -0,0 +1,83 @@
...
+    }
+    return true;
+  }
+  // TODO: Method that takes Map/Object as arg
+  // Endpoint on admin controller should call this.
+  // For now, just use that in other tests (since we can't modify this internal Map)
+  // So, call that from test e.g. map submission to set switch to true, check it 503s, bish bash bosh.
+
+  // On init, load entry (see nestjs lifecycle stuff for how) from config table in DB (lets call it Killswitches) and store on map,
+  // when killswitches are updated, update both js map and value in DB
```
tsa96, line 82: You can remove this now!

apps/backend/src/app/modules/killswitch/killswitch.service.ts
```diff
@@ -0,0 +1,83 @@
...
+      );
+    }
+  }
+
+  public checkKillswitch(type: KillswitchType): boolean {
+    return this.switches[type];
+  }
+
+  public async updateKillswitches(switches: Killswitches): Promise<void> {
+    this.validateKillswitches(switches);
```
tsa96, line 45: Same as above, we need to handle whether this returns false. Throwing a 400 bad request is fine.

tsa96 (changes requested):
@kanoka42 I started reviewing this just now but see you're still force-pushing.

I'll post what I currently have, but leave the rest until stuff is stable. Best to not mark as "ready for review" til you're happy though not a big deal at all, just lmk when you're ready.

One general comment is the commit message on the killswitch commit is a bit unclear, something like `feat(back): add killswitch endpoint guard system` just to make it clear this is adding a whole new system, "killswitch" feels a bit vague

apps/backend/src/app/modules/killswitch/killswitch.service.ts
kanoka42: So create a new logger instance in the service?

tsa96: Yeah, though tbh I don't remember exactly how we do it, have a look round the codebase

apps/backend/src/app/modules/killswitch/killswitch.service.ts
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  public checkKillswitch(type: KillswitchType): boolean {
+    return this.switches[type];
+  }
+
+  public async updateKillswitches(switches: Killswitches): Promise<void> {
+    const areValidSwitches = this.validateKillswitches(switches);
+
+    if (!areValidSwitches) {
```
tsa96, line 73: Just use `if (!this.validateKillswitches(switches)`. It's preference I guess, but I think creating named variables just to use once, on the immediate line below, tends to hurt readability.

apps/backend/src/app/modules/killswitch/killswitch.service.ts
```diff
@@ -0,0 +1,106 @@
...
+    const switchTypes = Enum.values(KillswitchType);
+    for (const [type, value] of Object.entries(switches)) {
+      if (
+        !(value === true || value === false) ||
+        !switchTypes.includes(type as KillswitchType)
+      ) {
+        return false;
+      }
+    }
+    return true;
```
tsa96, line 104: Oh just realised, this can just be 
```ts
return Object.entries(switches).every(([type, value]) => (value === true || value === false) && switchTypes.includes(type as KillswitchType))
```
(let prettier do the heavy lifting)

libs/db/src/prisma/schema.prisma
```diff
@@ -589,3 +589,11 @@ model AdminActivity {
...
   createdAt DateTime @default(now())
 }
+
+
+/// Not standard SQL practice but whatever, this is a solid way to persist
+/// config
+model Config {
+  id             String @id
+  value          Json
+}
```
tsa96, line 599: Missing newline at end of file. My fault, sorry!
Remember to `fixup` this commit, e.g. commit just the fix for this, then
```
git commit -m 'fixup schema'`
git rebase -i HEAD~2
```
then move that commit directly below the schema change commit and `fixup`. Or just use the Fork UI!

apps/backend-e2e/src/map-review.e2e-spec.ts
```diff
@@ -331,6 +334,41 @@ describe('Map Reviews', () => {
...
+          token: adminToken
+        });
+
+        const response = await req.del({
+          url: `map-review/${review.id}`,
+          status: 503,
+          token: token
+        });
+
+        expect(response.body.statusCode).toEqual(503);
```
tsa96, line 357: Unnecessary, the `status: 503` check in `req.del` is doing exactly this under the hood.

apps/backend-e2e/src/map-review.e2e-spec.ts
```diff
@@ -331,6 +334,41 @@ describe('Map Reviews', () => {
...
+          url: 'admin/killswitch',
+          status: 204,
+          body: {
+            NEW_SIGNUPS: false,
+            RUN_SUBMISSION: false,
+            MAP_SUBMISSION: false,
+            MAP_REVIEWS: false
+          },
+          token: adminToken
+        });
```
tsa96, line 369: I don't like repeating this so frequently, better to have a dedicated function for it
[website-e865e35-feat(back-e2e) util for resetting killswitches.patch](https://github.com/user-attachments/files/15749302/website-e865e35-feat.back-e2e.util.for.resetting.killswitches.patch)
wasn't sure where to put it, here's a patch adding a tmp function to request.util until we move this system to redis.

tsa96: Remember to update this on all the other tests!

apps/backend-e2e/src/map-review.e2e-spec.ts
tsa96: Applies to loads below

apps/backend/src/app/modules/admin/admin.controller.ts
```diff
@@ -63,9 +67,36 @@ export class AdminController {
...
 
+  @Patch('/killswitch')
+  @HttpCode(HttpStatus.NO_CONTENT)
+  @ApiOperation({ summary: 'Update active killswitches' })
+  @ApiNoContentResponse({
+    description: 'The switches were updated successfully'
+  })
+  @ApiBadRequestResponse({ description: 'Invalid switches type' })
+  @ApiBody({
+    type: Object,
```
tsa96, line 82: Fine to not do a schema for this (annoying to do) but maybe update description in line below to say e.g. `Object of type Record<KillswitchType, boolean>`

apps/backend/src/app/modules/admin/admin.controller.ts
```diff
@@ -63,9 +67,36 @@ export class AdminController {
...
+    this.killswitchService.updateKillswitches(switches).then((r) => r);
+  }
+
+  @Get('/killswitch')
+  @ApiOperation({ summary: 'Gets currently stored killswitches' })
+  @ApiNoContentResponse({
+    description: 'Found switches'
+  })
+  @ApiBadRequestResponse({ description: 'Invalid switches type' })
+  getKillSwitches(): object {
```
tsa96, line 96: Can't you use `Killswitches` instead of `object` here? And same thing for the ApiNoContentResponse as the ApiBody above, just a text-based type in the description in fine.

tsa96 (changes requested):
We should still add a test or two to unit tests in users.service.spec.ts

But really good work overall, done a great job familiarizing yourself given you're new to TypeScript and Nest!

apps/backend/src/app/modules/admin/admin.controller.ts
```diff
@@ -63,9 +67,36 @@ export class AdminController {
...
+  })
+  @ApiBadRequestResponse({ description: 'Invalid switches type' })
+  @ApiBody({
+    type: Object,
+    description: 'Object of type Record<KillswitchType, boolean>',
+    required: true
+  })
+  updateKillSwitch(@Body() switches: Killswitches): void {
+    this.killswitchService.updateKillswitches(switches).then((r) => r);
+  }
```
tsa96, line 88: `.then((r) => r)` have no effect whatever besides make ESLint think you're not ignoring a promise. `.then` behaves kind like `.map` on arrays, transforming whatever the promise returns after it resolves (if you want to be nerdy I *think* it's a monad). `.then(r => r)` says, "when this promise resolves, map `r` to `r` - has no absolutely no affect.

In this case you just want to `await` the Promise and not return it. That ensures the work being done inside the promise actually completes before we return a response, but we don't actually return whatever the promise resolves out to. In your case you're not actually awaiting anything, this endpoint will return as soon as `updateKillswitches` returns a promise, before the ongoing async work in that method has completed.

tsa96 (changes requested):
Almost there, just need e2e tests for new admin endpoints

apps/backend/src/app/modules/admin/admin.controller.ts
kanoka42: Updated in [91fe759](https://github.com/momentum-mod/website/commit/91fe759399519f03c2b85ca32f85968d4e0d5ea7)

tsa96:
I'm sorry, I mentioned adding a couple unit tests in users.service.spec.ts a couple comments ago then forgot to bring it up in my last review. Since we're already testing the findOrCreateUser method we should definitely be including a check for the killswitch guard in there.

Besides that everything looks good, will approve after those tests

apps/backend/src/app/modules/users/users.service.spec.ts
```diff
@@ -254,6 +259,94 @@ describe('UserService', () => {
       });
     });
 
+    it('should not create a new  if the killswitch is active', async () => {
+      // await killswitchService.onModuleInit();
```
tsa96, line 263: leftover testing comment?

tsa96 (changes requested):

tsa96 (approved):

PR #956, opened 2024-06-17
Add Healthcheck Endpoint, Bind Server to 0.0.0.0 for Prod

Gocnak:
It seems like by default fastify binds to 127.0.01 which is great for localhost development but not working in Prod. They mention it [in their getting started guide](https://fastify.dev/docs/latest/Guides/Getting-Started/#note), and while we do have `HOST` bound to `0.0.0.0` in the Dockerfile, I don't actually think Fastify was reading this?

Also, I used `fastify-healthcheck` to add a health check endpoint for Docker and our uptime page to get a proper status of the API. I'm not married to this specific implementation, we can always roll one ourselves, but this just seemed drag-n-drop and easy enough.

### DIFF
--- a/apps/backend/src/app/app.module.ts
+++ b/apps/backend/src/app/app.module.ts
@@ -23,6 +23,7 @@ import { ExtendedPrismaService } from './modules/database/prisma.extension';
 import { EXTENDED_PRISMA_SERVICE } from './modules/database/db.constants';
 import { DbModule } from './modules/database/db.module';
 import { KillswitchModule } from './modules/killswitch/killswitch.module';
+import { HealthcheckModule } from './modules/healthcheck/healthcheck.module';
 
 @Module({
   imports: [
@@ -92,7 +93,8 @@ import { KillswitchModule } from './modules/killswitch/killswitch.module';
     UsersModule,
     SessionModule,
     XpSystemsModule,
-    KillswitchModule
+    KillswitchModule,
+    HealthcheckModule
   ],
   providers: [
     {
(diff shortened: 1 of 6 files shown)


tsa96 (changes requested):
I'm not too keen on the healthcheck implementation. It's an extra dependency for something ***very*** simple.

The major change is using https://www.npmjs.com/package/@fastify/under-pressure, which we could use in the future, but if we probably want to make the settings configurable via env vars. If we want to do that as part of prod work I'm all for it, but I think we should do the healthcheck ourselves anyway. For now we're just trying to set up a docker healthcheck, so something that 204s for every request seems fine.

The source for this plugin is quite funny - if you strip away the configuration stuff it basically amounts to this:
```ts
 fastify.register(require('@fastify/under-pressure'), {});

 fastify.route({
      method: 'GET',
      url: '/health,
      handler: (req, reply) => reply.code(200).send({ statusCode: 200, status: 'ok' })
  });
```
https://github.com/smartiniOnGitHub/fastify-healthcheck/blob/master/src/plugin.js

so under-pressure does all the actual work.

So we can implement this healthcheck endpoint ourselves in the exactly same fashion and skip the dependency. I agree that doing it directly with Fastify rather than a separate NestJS module and controller is fine, and we shouldn't ever need more public healthchecks for anything besides this when we have metrics for all our internal services e.g. Postgres, Redis being sent to Grafana.

tsa96:
Actually ugh, I think this might start responding too early, before Nest has done all it's module resolution stuff. I hate having to do a whole module for such a simple thing but actually I think it's the best approach, going to change this, one sec.

tsa96:
Okay there's a version as Nest module

tsa96 (approved):

tsa96:
I'll let you merge once happy with my stuff

PR #993, opened 2024-07-06
Fix SentryInterceptor

tsa96:
SentryInterceptor should *hopefully* work now. Since this is TypeScript-specific DI stuff I can't test this by tweaking main.js in prod 🙃


tsa96:
Or rather, the interceptor definitely works, just haven't tried it out on prod yet

Gocnak (approved):

PR #1018, opened 2024-08-17
fix(build): Make backend user in docker own node_modules

GordiNoki:
Makes all files in /app folder of docker container be owned by backend user.
Should help with migration errors

### Checks

- [x] I have ran `nx run db:create-migration` and committed the migration if I've made DB schema changes

### DIFF
--- a/prod/Dockerfile
+++ b/prod/Dockerfile
@@ -10,11 +10,11 @@ COPY ../libs/db/src/prisma/ ./prisma/
 
 RUN chown -R backend:backend .
 
+USER backend
 RUN npm --omit=dev --ignore-scripts install
 RUN node_modules/.bin/prisma generate --schema ./prisma/schema.prisma
 
 RUN rm package.json package-lock.json
-COPY ./prod/package.json .
+COPY --chown=backend ./prod/package.json .
 
-USER backend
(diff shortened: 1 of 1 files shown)


prod/Dockerfile
```diff
@@ -10,11 +10,14 @@ COPY ../libs/db/src/prisma/ ./prisma/
...
 
+USER backend
 RUN npm --omit=dev --ignore-scripts install
 RUN node_modules/.bin/prisma generate --schema ./prisma/schema.prisma
 
 RUN rm package.json package-lock.json
 COPY ./prod/package.json .
 
+RUN chown -R backend:backend ./package.json
+
```
Gocnak, line 21: COPY actually has a `--chown` flag: https://docs.docker.com/reference/dockerfile/#copy---chown---chmod so this isn't needed if you do that

prod/Dockerfile
```diff
@@ -10,11 +10,14 @@ COPY ../libs/db/src/prisma/ ./prisma/
...
 USER backend
```
Gocnak, line 22: Would this USER down here now be redundant and can be removed?

Gocnak (changes requested):

Gocnak (approved):

Issue #916, opened 2024-03-01
Leaderboard Updating, Leaderboard Resetting, Run Deletion


tsa96:
[murgle started working on this a while ago](https://github.com/momentum-mod/website/pull/580), but it's for the old backend and frontend code is really outdated, so not much point keeping the PR open given it'd be best to just remake it anyway. Closing that one, and getting an issue started for the feature.

I'm a little hesitant to start working on this yet, since post 0.10 leaderboard and rank XP systems will completely change how we process and store leaderboard rankings and XP. So, leaving this blocks for now, but we should work on this once we start doing rank/leaderboard changes.

tsa96:
We should also add an endpoint for resetting entire leaderboards. This will be very rare and should only be available to admins.

tsa96:
Adding even more onto this issue. Currently admins/mods can't submit `MapVersion`s, but we definitely want this. However, if we allow zone updates in that case, we may have leaderboard updates to as well, and the logic seems very complicated.

I'd love to get this in for 0.10.0, but not sure it's worth working on right now.

tsa96:
Don't think blocking because of future ranking stuff makes sense, the current system is so simple and we don't even display rank XP anwhere, plus we're probably going to work on ranking very soon anyway, so if we completely break the current ranking stuff when doing this I don't think it really matters.

tsa96:
Assigning myself since I can only imagine myself or Gordi doing this, and he's busy for a while.

tsa96:
Lowering priority, not essential for map submission being "completed" and ready for people to start submitting stuff to, but we should definitely try to get this in 0.10.0

PR #1032, opened 2024-09-03
Generalized Map Versioning

### LINKED ISSUE #857: Don't include `zones` field on maps on all MMap SELECTs
The `zones` JSON field on the `MMap` table can be enormous, shouldn't be included in DB queries for/including MMap unless explicitly requested. This is quite annoying - we constantly do joins on other tables including MMap, plus Prisma's `MMapSelect` type of really ugly, needs booleans for every field you want to include.

Few methods I considered: 
1. Use an explicit `Prisma.MMapSelect` all over the place, probably include some kind of integration with the `expandToIncludes` function.
2. Split out to a `MapZones` table, 1-1 w/ `MMap`, using mapID as primary key. Simplifies code quite a bit, but an extra DB table.
3. Some Prisma Client Extension crap.

(2) is Prisma-agnostic but I'm not a fan of the extra table, so probably going with (1). Have a stash starting on this (started on this just now but realised isn't really necessary for ongoing PR) will pick it up in the future.

### LINKED ISSUE #1006: Allow admins to fast-track other user's maps from public testing to final approval
https://github.com/momentum-mod/website/pull/1004 did half of this, but mods don't actually have permission to change other user's map status. Again, I think we should be giving admins and mods free reign over the map submission pipeline - we know who everyone else and expect them to behave. This allows for cases where we want a fast-track a map a non-moderator submitted, if there's a reason to fast-track, e.g. for competitions, livestreams (fangames recently would've needed this).

tsa96:
Closes #1006
Closes #857

Replaces the `MapSubmissionVersion` table with `MapVersion`. Map storage is now always using the zone system, even for approved maps. Previously, prior to approval maps, would be stored at `submissions/<MapSubmissionVersion UUID ID>.bsp`, then upon approval moved to `maps/<map name>.bsp`. Now, all maps are just stored at `maps/<MapVersion UUID ID>.bsp`. This makes frontend and game code quite a lot simpler.

Whilst I haven't done admin map updating quite yet (https://github.com/momentum-mod/website/issues/916), this change is essential to that system.

Also, since zones are now always stored on MapVersion, so always requires a join table to fetch, it's a lot easier to avoid fetching the zones JSON unnecessarily, plus I added Prisma's preview "omit" API which simplifies code even more. So https://github.com/momentum-mod/website/issues/857 can be closed in the process. 

### Checks

- [x] I have ran `nx run db:create-migration` and committed the migration if I've made DB schema changes

### DIFF
--- a/apps/frontend/src/app/components/map-review/map-review-suggestions-form.component.ts
+++ b/apps/frontend/src/app/components/map-review/map-review-suggestions-form.component.ts
@@ -59,9 +59,7 @@ export class MapReviewSuggestionsFormComponent implements ControlValueAccessor {
         label: GamemodeInfo.get(gamemode).name
       }));
     this.availableBonusTracks =
-      (
-        map?.zones ?? map?.submission?.currentVersion?.zones
-      )?.tracks?.bonuses?.map((_, i) => ({
+      map?.currentVersion?.zones?.tracks?.bonuses?.map((_, i) => ({
         trackNum: i + 1,
         label: i + 1
       })) ?? [];
--- a/apps/frontend/src/app/components/map-review/map-review-suggestions-form.component.ts
+++ b/apps/frontend/src/app/components/map-review/map-review-suggestions-form.component.ts
@@ -92,7 +92,6 @@ export class MapReviewSuggestionsFormComponent implements ControlValueAccessor {
           .map(({ trackNum }) => trackNum)
       );
 
-      trackNum = 1;
       while (bonusNums.has(trackNum)) trackNum++;
     }
 
--- a/apps/frontend/src/app/pages/maps/map-info/map-info.component.html
+++ b/apps/frontend/src/app/pages/maps/map-info/map-info.component.html
@@ -34,7 +34,7 @@
           <m-icon class="mr-2 pb-0.5" icon="play-circle" />
           Play
         </button>
-        @if (map.downloadURL ?? map.submission?.currentVersion?.downloadURL; as bsp) {
+        @if (map.currentVersion.downloadURL; as bsp) {
           <a (click)="downloadOverlay.show($event)" class="btn btn-purple" mTooltip="Download map file">
             <m-icon class="pb-0.5" icon="download" />
             <p-overlayPanel #downloadOverlay appendTo="body">
@@ -48,7 +48,7 @@
                 <p>
                   <a class="link cursor-pointer" (click)="downloadZoneFile(map)">Download Zones</a>
                 </p>
-                @if (map.vmfDownloadURL ?? map.submission?.currentVersion?.vmfDownloadURL; as vmf) {
+                @if (map.currentVersion?.vmfDownloadURL; as vmf) {
                   <p><a class="link" [href]="vmf">Download VMF</a></p>
                 }
               </div>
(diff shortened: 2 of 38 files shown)


apps/frontend/src/app/components/map-review/map-review-suggestions-form.component.ts
```diff
@@ -94,12 +94,8 @@ export class MapReviewSuggestionsFormComponent implements ControlValueAccessor {
           .map(({ trackNum }) => trackNum)
       );
 
-      for (let i = 0; bonusNums.size < 0; i++) {
-        if (!bonusNums.has(i)) {
-          trackNum = i;
-          break;
-        }
-      }
+      trackNum = 1;
```
Panzerhandschuh, line 95: Is this assignment necessary? It's already assigned to 1 on line 84.

Panzerhandschuh (approved):

apps/frontend/src/app/pages/maps/map-info/map-info.component.html
```diff
@@ -34,7 +34,7 @@
           <m-icon class="mr-2 pb-0.5" icon="play-circle" />
           Play
         </button>
-        @if (map.downloadURL ?? map.submission?.currentVersion?.downloadURL; as bsp) {
+        @if (map.currentVersion; as bsp) {
```
GordiNoki, line 37: Either this should be `map.currentVersion.downloadURL`, or you should add `.downloadURL` on line 46

apps/frontend/src/app/components/search/user-search.component.html
```diff
@@ -16,7 +16,7 @@
   />
 </div>
 <ng-template #results>
-  <div class="flex min-h-[3.75rem] flex-col justify-center gap-2" [ngClass]="{ 'w-96': useOverlay }">
+  <div class="flex min-h-[3.75rem] flex-col justify-center [&>:not(:first-child)]:mt-2" [ngClass]="{ 'w-96': useOverlay }">
```
GordiNoki, line 19: What's the difference?

GordiNoki (changes requested):
I'm not sure, but I remember someone on engine side asking for map downloads to have the same file name as a map file that is being downloaded.

tsa96:
> I'm not sure, but I remember someone on engine side asking for map downloads to have the same file name as a map file that is being downloaded.

Nope not a problem for us, our HTTP downloader code just fetches a big ol' buffer and we can write it wherever.

apps/frontend/src/app/components/map-review/map-review-suggestions-form.component.ts
tsa96: Good spot, removing

tsa96: oh, this was a commit on a previous PR which I merged yesterday, whatever, will sneak a tiny commit into here

apps/frontend/src/app/pages/maps/map-info/map-info.component.html
tsa96: Good spot, fixed

apps/frontend/src/app/components/search/user-search.component.html
tsa96: There's a pagination control or something taking up space in when invisible, I don't remember exactly

GordiNoki: It doesn't fix it tho. It still adds margin to invisible paginator

tsa96: I'm not sure that's right, I worry that destroying the paginator might lose some important state. But since this is unclear let's not slow the PR down with something unrelated, going to ditch the commit. I don't even think this is a significant issue for a follow-up issue.

GordiNoki (approved):

tsa96:
Not a big deal in this case but would prefer you let me merge my PRs @GordiNoki, at least when it's backend changes - when we have corresponding changes in red it's easiest if they all get in at once. If it's something you're waiting on feel free to badger me or nick

PR #1035, opened 2024-09-09
feat(front): implement level indicator

### LINKED ISSUE #870: Level indicator Angular component
Panorama has a small component for showing user level and prestige icon. It's relatively simple, and all open source. We should do an identical component in Angular.

https://github.com/momentum-mod/panorama/blob/master/layout/components/level-indicator.xml
https://github.com/momentum-mod/panorama/blob/master/scripts/components/level-indicator.js
https://github.com/momentum-mod/panorama/blob/master/styles/components/level-indicator.scss

Animations don't matter, they're only used for when the player levels up in real-time ingame. What's important is that we get the same colours and prestige icons.

Note that whilst the colours used aren't currently in our website repo or Tailwind config, we definitely want them in the future. I'm not 100% sure if they'll be the base of full colour palettes (that would be nice), but I'll get them organised in the new [*styling* repo](https://github.com/momentum-mod/styling) in the near future. If I haven't done that by the time someone else picks this up, nag me about it.

Here's an image for reference, though I assume anyone working on this probably has access on Steam already, but if not, just contact me for one.
Momentum_Mod_-_DX1105,12,23,103505

pedroalves80:
Closes #870 

Implemented level indicator with prestige in new component, same as panorama.
Will need some tweaks to use tailwind bg classes in the future when the new tailwind palettes are finished as @tsa96 said.

image
image
image
image

### Checks

- [ ] I have ran `nx run db:create-migration` and committed the migration if I've made DB schema changes
- [] All changes requested in review have been `fixup`ed into my original commits


apps/frontend/src/app/components/level-indicator/level-indicator.component.ts
```diff
@@ -0,0 +1,53 @@
...
+  protected prestigeIcon: string;
+  protected textColor: string;
+  protected imageClass: string;
+
+  @Input()
+  totalLevel: number;
+
+  constructor(private readonly xpService: XpSystemsService) {}
+
+  ngOnInit() {
```
tsa96, line 24: `ngOnInit` isn't right here, both because the local user service might not have fetched the latest user data when this component is first created, and because the user data might update at some point. User's can't change their XP just by using the site, but we could have cases where they're playing the game at the time as having the site open, and we'd want that to affect this component.

Again it's Angular so there's several ways to do this, but simplest in this case is to just replace `OnInit`/`ngOnInit` with `OnChanges`/`ngOnChanges`. You only have one input, so if that method is called, the level must have updated.

apps/frontend/src/app/components/level-indicator/level-indicator.component.ts
```diff
@@ -0,0 +1,53 @@
...
+      { max: 150, color: '#ff8645' },
+      { max: 200, color: '#ffd000' },
+      { max: 250, color: '#47c27c' },
+      { max: 300, color: '#3f4aca' },
+      { max: 350, color: '#873de1' },
+      { max: 400, color: '#d34a8d' },
+      { max: 450, color: '#3185fc' },
+      { max: 500, color: '#161616' },
+      { max: Infinity, color: '#ffffff' }
+    ];
```
tsa96, line 49: Not that we really care about perf here, but I'd still move this to a `private static readonly` field of this class so you're not creating this same array every time this method is called.

apps/frontend/src/app/components/level-indicator/level-indicator.component.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    return levelColors.find((l) => level < l.max)?.color || '#5d5d5d';
```
tsa96, line 51: Can do `find(({max}) => level < max)` to say the random `l`. Also, better to use `??` than `||` here - though I don't see the need to use either, if `level` in a `number` then how could this `levelColors.find` ever be nullish?

apps/frontend/src/app/components/level-indicator/level-indicator.component.html
```diff
@@ -0,0 +1,13 @@
+<div
+  [ngStyle]="{ 'background-color': levelColor }"
+  [ngClass]="[prestige ? 'pl-1 pr-3' : 'px-2', textColor]"
+  class="flex gap-1 rounded py-0 font-display text-xl shadow"
```
tsa96, line 4: How does bold text here look? IIRC that's what we do ingame, and looks a little thin in your screenshots.

libs/xp-systems/src/xp-systems.class.ts
```diff
@@ -187,6 +187,20 @@ export class XpSystems {
...
+  getInnerLevel(totallevel: number): number {
+    return ((totallevel - 1) % 500) + 1;
+  }
+
+  getPrestigeIcon(prestige: number): string {
+    return prestige <= 5
+      ? `assets/images/prestige/prestige${prestige}.svg`
+      : 'assets/images/prestige/max_level.svg';
+  }
+
```
tsa96, line 203: This code lives in a shared library so would rather not have frontend-specific paths in here - I'd just stick this method inside of the level indicator component

tsa96 (changes requested):

tsa96:
Thanks for this! One other thing I just noticed, the text in the level indicator looks like it's a few pixels lower than the icon, are you sure it's centered right?

apps/frontend/src/app/components/level-indicator/level-indicator.component.html
pedroalves80: Using bold looks like this: 
image

As ingame looks like this:
image

Looks more closer to ingame using bold, but font family is still different I think.

pedroalves80:
> Thanks for this! One other thing I just noticed, the text in the level indicator looks like it's a few pixels lower than the icon, are you sure it's centered right?

From what I've seen, Is correctly centered. Maybe just depends on the perspective of the  number: 
image

apps/frontend/src/app/components/level-indicator/level-indicator.component.ts
pedroalves80: You're right, removed that condition.

apps/frontend/src/app/components/level-indicator/level-indicator.component.html
pedroalves80: Added b tag to level number.

tsa96:
Okay, LGTM. Thanks for this!!

tsa96 (approved):

PR #1050, opened 2024-10-14
ci: separate staging.yml for frontend and backend

1WHISKY:
Separate staging.yml to staging_frontend.yml and staging_backend.yml
This allows us to use the "paths" parameter, so that changes to the frontend only run the frontend action and changes to the backend only run the backend action instead of running both.

### Checks

### DIFF
--- a/.github/workflows/staging_backend.yml
+++ b/.github/workflows/staging_backend.yml
@@ -1,0 +1,16 @@
+name: Publish Staging Backend
+
+on:
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - "apps/frontend*/**"
+      - "scripts/**"
+
+jobs:
+  publish_backend:
+    permissions:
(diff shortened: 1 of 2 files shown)


tsa96 (changes requested):
This change is seems okay, though I'm really not concerned about us re-deploying staging too often - it doesn't cost us anything, and brief staging outages are not a problem.

@Gocnak you should probably take a look at this as well.

tsa96:
Yeah. Staging deploys don't cost us anything and I'd much rather we run these jobs when not necessary, than fail to run them when we should. We could exclude other files, but I don't want a maintenence burden of updating the that list.

tsa96 (approved):

tsa96:
Waiting for Nick to check this out, might take a bit, but this isn't high priority.

PR #1062, opened 2024-10-23
Fix cosmetic xp calculations

### LINKED ISSUE #1045: User XP defaulting to -21000
I thought I fixed this but I guess not. Might not have noticed it in dev, since the seed script usually creates an admin user itself, with the correct XP.

Don't have time to look into into at all, but shouldn't be hard to figure out. Could try just stepping through user creation steps with a debugger.

To test this, try starting up both front and backend and logging in *without* running the seed script.

tsa96:
Closes #1045

Changes the code precalculating XP values in xp-systems to match [C++](https://github.com/momentum-mod/game/blob/develop/mp/src/game/shared/momentum/util/mom_system_xp.cpp). We don't unit test the C++ code, but I've manually checked all the expected values in our unit tests with the C++ version using a debugger.

After this change, the frontend components have the correct values (0 / 21000 XP for new user at level 1).

I also improved the docs for these functions (which were partially the cause of the confusion here) and cleaned up the xp-systems class a bit.

### Checks

### DIFF
--- a/apps/frontend/src/app/components/level-indicator/level-indicator.component.html
+++ b/apps/frontend/src/app/components/level-indicator/level-indicator.component.html
@@ -1,11 +1,12 @@
 <div
   [ngStyle]="{ 'background-color': levelColor }"
-  [ngClass]="[prestige ? 'pl-1 pr-3' : 'px-2', textColor]"
-  class="flex items-center gap-1 rounded py-0 font-display text-xl shadow"
+  [ngClass]="[textColor, prestige ? 'pl-1' : 'pl-2']"
+  class="flex items-center gap-1 rounded pr-2 font-display shadow"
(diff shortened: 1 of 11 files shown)


GordiNoki (approved):
Nice work, but I think we should also fix seed script, since now it generates users with negative xp:
```ts
userStats: {
  create: {
    totalJumps: Random.int(10000),
    totalStrafes: Random.int(10000),
    level: Random.int(0, 1000), // I don't think this should include 0 as min value, since sure, it won't ever be 0, but it's not a valid level anyway
    cosXP: Random.int(10000), // Most levels would have much more xp required for a level, so I think this one should be calculated as xp required for current level plus some random value with max being xp to get to next level
    mapsCompleted: Random.int(10000),
    runsSubmitted: Random.int(10000)
  }
}
```
изображение

tsa96:
Yup, will do

PR #1082, opened 2024-12-11
feat: throw error if duplicate map

### LINKED ISSUE #245: Hash the map on the backend and compare against other map hashes to prevent duplicate file uploading

jef:
Closes #245

- Tries to find if the hash of the .bsp is in the database. If it is, then it throws a `ConflictException`.

### Checks

### DIFF
--- a/apps/backend/src/app/modules/maps/maps.service.ts
+++ b/apps/backend/src/app/modules/maps/maps.service.ts
@@ -671,13 +671,13 @@ export class MapsService {
     const bspHash = FileStoreService.getHashForBuffer(bspFile.buffer);
 
     // Check for duplicate map hash
-    // const existingMap = await this.db.mMap.exists({
-    //   where: { versions: { some: { bspHash } } }
-    // });
+    const existingMap = await this.db.mMap.exists({
+      where: { versions: { some: { bspHash } } }
+    });
 
-    // if (existingMap) {
-    //   throw new ConflictException('Map with this file hash already exists');
-    // }
+    if (existingMap) {
+      throw new ConflictException('Map with this file hash already exists');
+    }
 
     let zonesStr: string;
     let zones: MapZones;
--- a/apps/backend/src/app/modules/maps/maps.service.ts
+++ b/apps/backend/src/app/modules/maps/maps.service.ts
@@ -671,13 +671,13 @@ export class MapsService {
     const bspHash = FileStoreService.getHashForBuffer(bspFile.buffer);
 
     // Check for duplicate map hash
-    const existingMap = await this.db.mMap.exists({
-      where: { versions: { some: { bspHash } } }
-    });
+    // const existingMap = await this.db.mMap.exists({
+    //   where: { versions: { some: { bspHash } } }
+    // });
 
-    if (existingMap) {
-      throw new ConflictException('Map with this file hash already exists');
-    }
+    // if (existingMap) {
+    //   throw new ConflictException('Map with this file hash already exists');
+    // }
 
     let zonesStr: string;
     let zones: MapZones;
--- a/apps/backend/src/app/modules/maps/maps.service.ts
+++ b/apps/backend/src/app/modules/maps/maps.service.ts
@@ -552,6 +552,15 @@ export class MapsService {
     const hasVmf = vmfFiles?.length > 0;
     const bspHash = FileStoreService.getHashForBuffer(bspFile.buffer);
 
+    // Check for duplicate map hash
+    const existingMap = await this.db.mMap.exists({
+      where: { versions: { some: { bspHash } } }
+    });
+
+    if (existingMap) {
+      throw new ConflictException('Map with this file hash already exists');
+    }
+
     let map: Awaited<ReturnType<typeof this.createMapDbEntry>>;
 
     const tasks: Promise<any>[] = [
@@ -660,6 +669,15 @@ export class MapsService {
 
     const hasVmf = vmfFiles?.length > 0;
     const bspHash = FileStoreService.getHashForBuffer(bspFile.buffer);
+
+    // Check for duplicate map hash
+    const existingMap = await this.db.mMap.exists({
+      where: { versions: { some: { bspHash } } }
+    });
+
+    if (existingMap) {
+      throw new ConflictException('Map with this file hash already exists');
+    }
 
     let zonesStr: string;
     let zones: MapZones;
--- a/apps/backend-e2e/src/maps.e2e-spec.ts
+++ b/apps/backend-e2e/src/maps.e2e-spec.ts
@@ -1230,6 +1230,30 @@ describe('Maps', () => {
           });
         });
 
+        it('should throw 409 Conflict if map with same BSP hash is submitted again', async () => {
+          await uploadBspToPreSignedUrl(bspBuffer, token);
+
+          await req.postAttach({
+            url: 'maps',
+            status: 201,
+            data: createMapObject,
+            token
+          });
+
+          await uploadBspToPreSignedUrl(bspBuffer, token);
+
+          const res = await req.postAttach({
+            url: 'maps',
+            status: 409,
+            data: createMapObject,
+            token
+          });
+
+          expect(res.body.message).toBe(
+            'Map with this file hash already exists'
+          );
+        });
+
         it('should succeed if VMF file is missing', async () => {
           await uploadBspToPreSignedUrl(bspBuffer, token);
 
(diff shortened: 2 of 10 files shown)


apps/backend/src/app/modules/maps/maps.service.ts
```diff
@@ -552,6 +552,15 @@ export class MapsService {
     const hasVmf = vmfFiles?.length > 0;
     const bspHash = FileStoreService.getHashForBuffer(bspFile.buffer);
 
+    // Check for duplicate map hash
+    const existingMap = await this.db.mMap.findUnique({
+      where: { bspHash }
+    });
```
tsa96: Bit of a nitpick but we have the `.exists` client exists for this. Also prefer not using `findUnique` when we don't actually have a DB-level uniqueness constraint on `bspHash`.

apps/backend/src/app/modules/maps/maps.service.spec.ts
```diff
@@ -103,7 +103,7 @@ describe('MapsService', () => {
       pass
         ? expect(await service.getMapAndCheckReadAccess({ mapID: map.id, userID })).toMatchObject(map)
         : await expect(service.getMapAndCheckReadAccess({ mapID: map.id, userID })).rejects.toThrow(ForbiddenException);
-    
+
```
tsa96, line 112: Random auto-formatting, don't want it in commits, I suggest just using Prettier and disabling any other formatters. I'm not sure why Prettier doesn't handle this, but if it doesn't, then so be it - would rather let Prettier handle this exclusively and avoid other autoformatter changes than have random whitespace changes like this.

apps/backend/src/app/modules/maps/maps.service.ts
```diff
@@ -553,8 +553,8 @@ export class MapsService {
     const bspHash = FileStoreService.getHashForBuffer(bspFile.buffer);
 
     // Check for duplicate map hash
-    const existingMap = await this.db.mMap.findUnique({
-      where: { bspHash }
+    const existingMap = await this.db.mMap.findFirst({
+      where: { versions: { some: { bspHash } } }
```
tsa96, line 557: Again, can be `.exists`. Looks like this can be fixuped into the above commit?

package.json
```diff
@@ -149,5 +149,8 @@
     "type-fest": "4.25.0",
     "typescript": "5.4.5",
     "webpack-bundle-analyzer": "4.10.2"
+  },
+  "volta": {
+    "node": "20.18.1"
```
tsa96, line 154: Wrong commit, let's stick this and the CI change together. I'll discuss Volta in general in overall review comment

apps/backend/src/app/modules/maps/maps.service.spec.ts
```diff
@@ -74,27 +80,27 @@ describe('MapsService', () => {
...
     // otherwise should fail. Below code performs each check against each state.
     const tests = {
       [MapStatus.APPROVED]: 'any',
       [MapStatus.PUBLIC_TESTING]: 'any',
       [MapStatus.PRIVATE_TESTING]: ['admin', 'moderator', 'submitter', 'acceptedRequest', 'inCredits'],
-      [MapStatus.CONTENT_APPROVAL]: ['admin', 'moderator', 'submitter' ,'reviewer',  'acceptedRequest', 'inCredits'],
+      [MapStatus.CONTENT_APPROVAL]: ['admin', 'moderator', 'submitter', 'reviewer', 'acceptedRequest', 'inCredits'],
       [MapStatus.FINAL_APPROVAL]: 'any',
       [MapStatus.DISABLED]: ['admin', 'moderator']
     };
```
tsa96, line 106: Formatting again, same principle as previous comment, Prettier is good enough, doesn't help to have multiple autoformatters running at once.

apps/backend/src/app/modules/maps/maps.service.spec.ts
```diff
@@ -206,4 +212,41 @@ describe('MapsService', () => {
...
+        currentVersionID: '123',
+        createdAt: new Date(),
+        updatedAt: new Date()
+      });
+
+      await expect(service.submitMap(dto, userID, [])).rejects.toThrow(
+        ConflictException
+      );
+    });
+  });
```
tsa96, line 251: This should be an E2E test. They're a bit more complicated to setup I'm afraid (and admittedly we E2E test too much), but all the related tests are done as E2E, should be easy to fit hash check stuff in there.

apps/backend/src/app/modules/maps/maps.service.ts
```diff
@@ -760,7 +760,7 @@ export class MapsService {
     );
   }
 
-  private async checkCreateDto(userID: number, dto: CreateMapDto) {
+  async checkCreateDto(userID: number, dto: CreateMapDto) {
```
tsa96, line 763: Above tests being E2E means you shouldn't need this, in general don't like changing the access modifiers in service logic just so tests can access it!

tsa96 (changes requested):
Thanks for first contribution! This check makes sense, though prefer `.exists` (see comments). Though we actually have two different endpoints that handle BSP uploads, the initial submission, and updated versions after then. So you need a check in `MapsService.submitMapVersion` as well, and E2E tests there as well.

The Volta thing I'm okay with, I'd been meaning to get round to pinning an Node version at some point anyway. Personally I'm using NVM but fine with switching.

apps/backend/src/app/modules/maps/maps.service.spec.ts
jef: I think this one is removing the extra spaces.

apps/backend/src/app/modules/maps/maps.service.ts
jef: This is outdated and I started using `findFirst`. Happy to use `.exists` though!

apps/backend/src/app/modules/maps/maps.service.spec.ts
jef: I believe prettier did take care of this!

tsa96: This formatting is wrong but there's a `prettier-ignore` comment above a parent block so shouldn't be touched. Looking at the code it looks unnecessary and I don't remind removing the comment, though little concerned if your editor/whatever CLI command you're using is ignoring this

apps/backend/src/app/modules/maps/maps.service.spec.ts
tsa96: Oh you're right, it's the same issue with // prettier-ignore being ignored

apps/backend/src/app/modules/maps/maps.service.spec.ts
jef: Gotcha, okay I'll be more weary of what it's up to. I thought I just had Prettier running, but in either case, I'll revert all these changes in this file and put it in a E2E :)

tsa96: Are you using a Prettier extension, just letting commit hooks format stuff, or something else? Curious as to what's going on here.

tsa96:
Something weird's happened with Git here - if you need a hand just ping me on Discord in #website!

tsa96:
Just to clarify. might be that you just haven't been available over past few days, but in case you missed it, you seem to have botched the commits here, most stuff is now missing and we wouldn't accept a merge commit. No rush with the PR though!

jef:
> Just to clarify. might be that you just haven't been available over past few days, but in case you missed it, you seem to have botched the commits here, most stuff is now missing and we wouldn't accept a merge commit. No rush with the PR though!

Sorry, yes, I've been AFK and before I left only put the implementation. I will update the E2E when I get the _oomph_ again :D

tsa96:
All good, no rush

jef:
So I wrote up a test for at least `submitMap()` (I'll write up another for `submitMapVersion()` after successful), but I guess my code is causing another e2e test to fail...

tsa96:
Can't tell what one cus on phone, but one of these is missing e2e test!

jef:
The tests break more seemingly. I don't know if my logic is right, but it seems simply enough. I guess I don't know these endpoints good enough to implement the logic. I might give up on it for now.

tsa96:
> The tests break more seemingly. I don't know if my logic is right, but it seems simple enough. I guess I don't know these endpoints good enough to implement the logic. I might give up on it for now.

No problem, trying to debug E2E in a codebase you're unfamiliar with is never going to be fun. Appreciate the work you've done so far, I'll leave the PR open for now and finish it myself at some point, unless you decide to come back to it before then. Thanks for the contribution!

PR #1085, opened 2024-12-15
MomTV Run Submission

tsa96:
Reworks run submission to use new replay format including significantly refactoring validation logic. I've tested fairly extensively in-game, all working well.

Keen that this is reviewed by game devs, especially @jason-e. Most important thing is `run-processor.class.ts` which handles validation, and stuff in `libs/formats/replay`.

### Checks


apps/backend/src/app/modules/session/run/run-processor.class.ts
```diff
@@ -1,97 +1,130 @@
...
+import * as ReplayFile from '@momentum/formats/replay';
 import { CompletedRunSession, ProcessedRun } from './run-session.interface';
+import { approxEq } from '@momentum/util-fn';
+import { Logger } from '@nestjs/common';
 
 /**
  * Class for managing the parsing of a replay file and validating it against
  * run data
+ *
+ * I'm tempted to add a more advanced logging system into this that catpures
```
GordiNoki, line 20: Typo: cat purrs

GordiNoki (approved):
Looks great! I'm not completly sure about the run validation logic so I'll leave it to rio

apps/backend/src/app/modules/session/run/run-processor.class.ts
tsa96: Meow

apps/backend/src/app/modules/session/run/run-processor.class.ts
```diff
@@ -154,159 +191,176 @@ export class RunProcessor {
...
-      };
-    } catch {
-      throw new RunValidationError(ErrorType.BAD_REPLAY_FILE);
+    if (header.playerSteamID !== this.user.steamID)
+      throw new RunValidationError(ErrorType.BAD_META);
+
+    if (header.gamemode !== session.gamemode)
+      throw new RunValidationError(ErrorType.BAD_META);
+
+    if (!approxEq(header.tickInterval, Tickrates.get(session.gamemode)))
```
jason-e, line 270: Tick interval should be exactly the official value I would think

apps/backend/src/app/modules/session/run/run-processor.class.ts
```diff
@@ -154,159 +191,176 @@ export class RunProcessor {
...
+    // v runStart                              v header.timestamp    v now
+    // |---------------------------------------|---------------------|
+    // <--------------------------------------><--------------------->
+    //                    ^ headerRunTime                ^ submitDelay
+
+    // Duration of the run (double, in seconds!)
+    const headerRunTime = header.runTime * 1000;
+
+    // When the run was started according to replay file
+    const runStart = header.timestamp - headerRunTime;
```
jason-e, line 292: It's not an issue yet but I wanted to note something interesting to keep in mind: some of our movement code improvements involve stopping a movement tick short of completing the full tick, and in the future (probably coinciding with sub-tick timing in general), the timer will reflect these shortened ticks by ticking less than a full tick of time. This means that the end of run true timestamp minus the start of run true timestamp will likely be *slightly* more than the run duration. For most runs this will likely never matter, though it could amount to maybe even a few minutes difference for a 10 hour run. If needed, we could add to the replay data either a record of the tick adjustments or the actual run start timestamp.

apps/backend/src/app/modules/session/run/run-processor.class.ts
```diff
@@ -154,159 +191,176 @@ export class RunProcessor {
...
     // start zone.
-    if (timestamps[0].checkpoint !== 0) {
+    if (timestamps[0].minorNum !== 1) {
       throw new RunValidationError(ErrorType.BAD_TIMESTAMPS);
     }
 
     if (checkpointsOrdered) {
       for (let i = 1; i < timestamps.length; i++) {
-        if (timestamps[i].checkpoint <= timestamps[i - 1].checkpoint) {
+        if (timestamps[i].minorNum <= timestamps[i - 1].minorNum) {
```
jason-e, line 202: I think there also needs to be a check for all unique `minorNum` when `checkpointsOrdered === false`

apps/backend/src/app/modules/session/run/run-processor.class.ts
```diff
@@ -1,97 +1,130 @@
...
       throw new RunValidationError(ErrorType.BAD_TIMESTAMPS);
     }
 
-    // Check time is always increasing
-    for (let i = 1; i < this.timestamps.length; i++) {
-      if (this.timestamps[i].time < this.timestamps[i - 1].time) {
-        throw new RunValidationError(ErrorType.BAD_TIMESTAMPS);
-      }
+    // First minorNum is always 1 (a segment start), regardless of track type
+    if (timestamps[0].minorNum !== 1) {
```
jason-e, line 90: Is `timestamps` guaranteed to be ordered at this point, or could it still be out of order depending on network conditions? There's a few places where `timestamp` elements are accessed according to an assumed order.

jason-e (changes requested):
I only reviewed run-processor.class.ts based on your suggestion. Let me know if there's anything else I should review.

libs/constants/src/types/models/models.ts
```diff
@@ -523,6 +523,58 @@ export interface RunSessionTimestamp {
...
 
+export interface RunSplits {
+  trackStats: RunStats;
+  segments: RunSegment[];
+}
+
+export interface RunSegment {
+  subsegments: RunSubsegment[];
+
+  // Contains an entry for every subsegment the player has reached so far
```
jason-e, line 515: This comment should be in front of `subsegments` (the original C++ has the same error lol)

libs/constants/src/types/models/models.ts
```diff
@@ -523,6 +523,58 @@ export interface RunSessionTimestamp {
...
+  minorNum: uint8;
+
+  timeReached: float;
+
+  // Velocity when triggering this checkpoint; note the difference between this
+  // and Segment::effectiveStartVelocity
+  velocityWhenReached: vec3;
+}
+
+export interface RunStats {
```
jason-e, line 549: (nit) I can see you've just taken the original C++ class names and prepended "Run" to namespace them better, but it makes this one a little misleading since these stats can apply to not just a whole run but sometimes a whole segment of a run or a single subsegment.

apps/backend/src/app/modules/session/run/run-processor.class.ts
jason-e: If this is a matter of float32 vs JavaScript's native doubles, `Tickrates` should be defined as float32.

On a side note I just realized it's called `Tickrates` but stores tick *intervals* (the inverse of tickrate). I think we have a bad habit of using the terms interchangeably in game code too.

libs/formats/replay/src/replay-reader.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+        playerName: 'fingerprince',
+        tickInterval: expect.any(Number),
+        trackType: TrackType.MAIN,
+        trackNum: 1,
+        runTime: 30.584999316371977
+      };
+
+      const header = readHeader(buffer);
+      expect(header).toEqual(expectedHeader);
+      expect(header.tickInterval).toBeCloseTo(0.01, 5);
```
jason-e, line 36: Tick interval should be exact (making sure to compare float32 to float32 so no wiggle room is needed, not sure how easy/hard this is in JavaScript).

jason-e (changes requested):

apps/backend/src/app/modules/session/run/run-processor.class.ts
jason-e: If this is a matter of float32 vs JavaScript's native doubles, `Tickrates` should be defined as float32.

On a side note I just realized it's called `Tickrates` but stores tick *intervals* (the inverse of tickrate). I think we have a bad habit of using the terms interchangeably in game code too.

tsa96: Didn't know there was a nice way to do this with native JS until now, but turns out [`Math.fround`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/fround) is a thing and does what we're after.

My bad wrt the tickrate/intervals thing, definitely something we should be using precisely, have updated the naming.

libs/formats/replay/src/replay-reader.spec.ts
tsa96: Turns out, way easier than I initially though! Using `Math.fround(0.01)` instead of `0.01` passes exact equality.

tsa96:
Addressed a couple of rio's things but haven't organised commits yet or addressed other stuff, will do tomorrow

apps/backend/src/app/modules/session/run/run-processor.class.ts
tsa96: Duplicate `minorNum`s should *never* be possible, so I check for them `validateSessionTimestamps` already. Added a test for the case you give to make sure, which already passes.

libs/constants/src/types/models/models.ts
tsa96: Sure, can use a `RunSplits` namespace here.

apps/backend/src/app/modules/session/run/run-processor.class.ts
tsa96: Will refactor as discussed on Discord, but need to do when more awake (next couple days)

tsa96:
Addressed everything besides the timestamp ordering stuff which I'll do soon

apps/backend/src/app/modules/session/run/run-processor.class.ts
tsa96: I'm now performing a sort above here to ensure timestamps are first in the right order

jason-e (approved):

apps/backend/src/app/modules/session/run/run-processor.class.ts
```diff
@@ -122,20 +130,21 @@ export class RunProcessor {
     }
 
     // Main track runs from here on out.
-    // Segments are always ordered and required!
+    // Segments are always ordered NOOOO and required!
```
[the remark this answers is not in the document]
tsa96: Woops!!!

PR #1086, opened 2024-12-20
Map Tags

### LINKED ISSUE #503: Add tags to maps
Maps should have a configurable list of tags that can be set during the upload process. These allow for easier sorting and filtering of maps that match a specific description. The primary function of these is coordination with the learn section, listing maps that make heavy use of a particular technique that a player just learned how to do in the learn section (wallshots, speedshots, pogo, whatever). This can potentially be expanded later to also include the type of map (speedrun-oriented, linear, staged, Very Hard, whatever) but that is to be decided.

See the most recent public map selector mockup, above the map's leaderboard you can see a list of tags and how they will be used. (note: we will likely not be using aesthetic tags like these, those were just placeholders for the mockup design)
image

tsa96:
**Draft for now, frontend needs a little more work, just want to check CI**

Closes #503

Adds a tags field to leaderboards/leaderboard suggestions, an array of gamemode-dependent strings. 

Deliberately extremely simple, just plain JS strings stored in Postgres as a `text` array. Tags are stored in [TypeScript source](https://github.com/momentum-mod/website/blob/feat/map-tags/libs/constants/src/consts/map-tags.const.ts). Game C++ will never need to store individual tags, we'll just pass strings in from Panorama Typescript, which has a copy of that TS file via the fetch script that pulls everything from `libs/constants`.

I'm adding stuff according to [this google sheet](https://docs.google.com/spreadsheets/d/1iSaqxmobH70rf35AknE_yOWaJHnvksRIv-02qmS2a6M/edit?gid=0#gid=0), can continue adding new tags as people request stuff.

### Screenshots
image

### Checks


libs/constants/src/maps/map-tags.map.ts
tsa96, line 1: Not a fan of separate the enum file from everything else, I think you change this @GordiNoki? Easier to keep all in one place, would be fine having everything in `libs/constants/src/map-tags.ts`

GordiNoki (approved):

tsa96:
Nvm keeping the split stuff in `enums` and `maps` as-is since works well with Panorama import script already

PR #1098, opened 2025-02-25
feat(shared): possibility to follow steam friends

### LINKED ISSUE #897: Import Steam friends into follows
Steam's API lets us fetch all of a user's Steam friends, so we can use that to automatically follow them.

## Backend
Let's do a `GET /user/steamfriends` endpoint that uses the `SteamService.getSteamFriends` method to fetch Steam friends, then query DB for the matching SteamIDs. Then return an array of `UserDto` for every found user. A regular array is probably better than `PagedResponseDto` since we want to avoid calling Steam's API too often, and users can have at most 500 Steam friends. The likelihood that someone has more than ~100 friends with Momentum is always going to be pretty low.

Be sure to check out `LeaderboardRunsService.getRuns` to see an example of how to handle Steam friend data.

Then we'll want an endpoint for bulk following users. This can be at `POST /user/follow` and take an array (just CSV in the URL) of Momentum User IDs.

## Frontend 
Add a button with a Steam icon to the activity feed component that opens a new PrimeNG dialog component. It should contain a scrolling list of every user (`<a> containing their avatar and alias`) next to a checkbox, plus "Select All" and "Deselect All" buttons at the top right of the list. If >0 users are selected, enable a "Follow users" button that creates a `POST /user/follow` query on press, then closes the dialog.

pedroalves80:
Closes #897 

Created endpoints for steam friends and multiple user follows;
Added dialog modal to frontend, so the user can select multiple friends to follow;
Changed avatar size in search to match dialog avatar;

image
image
image
image
image

### Checks

### DIFF
--- a/apps/backend/src/app/modules/users/users.service.ts
+++ b/apps/backend/src/app/modules/users/users.service.ts
@@ -480,6 +480,41 @@ export class UsersService {
     return DtoFactory(FollowDto, dbResponse);
   }
 
+  async followUsers(userID: number, usersIDs: number[]): Promise<FollowDto[]> {
+    const users = await this.db.user.findMany({
+      where: { id: { in: usersIDs } }
+    });
+
+    if (users.length !== usersIDs.length)
+      throw new NotFoundException('One or more users not found');
+
+    if (usersIDs.includes(userID)) {
+      throw new BadRequestException('User cannot follow themselves');
+    }
+
+    const existingFollows = await this.db.follow.findMany({
+      where: {
+        followeeID: userID,
+        followedID: { in: usersIDs }
+      }
+    });
+
+    if (existingFollows.length > 0) {
+      throw new BadRequestException(
+        'User is already following one or more of the target users'
+      );
+    }
+
+    const createdFollows = await this.db.follow.createManyAndReturn({
+      data: usersIDs.map((followedID) => ({
+        followeeID: userID,
+        followedID
+      }))
+    });
+
+    return createdFollows.map((follow) => DtoFactory(FollowDto, follow));
+  }
+
   async updateFollow(
     localUserID: number,
     targetUserID: number,
@@ -745,6 +780,20 @@ export class UsersService {
     });
 
     return new PagedResponseDto(MapCreditDto, dbResponse);
+  }
+
+  async findSteamFriends(userID: number, steamID: bigint): Promise<UserDto[]> {
+    const steamFriends = await this.steamService.getSteamFriends(steamID);
+    const steamIDs = steamFriends.map((friend) => BigInt(friend.steamid));
+
+    const users = await this.db.user.findMany({
+      where: {
+        steamID: { in: steamIDs },
+        followers: { none: { followeeID: userID } }
+      }
+    });
+
+    return users.map((user) => DtoFactory(UserDto, user));
   }
 
   //#endregion
--- a/apps/frontend/src/app/pages/profile/profile.component.html
+++ b/apps/frontend/src/app/pages/profile/profile.component.html
@@ -154,29 +154,98 @@
           }
         </div>
 
-        <div [mSpinner]="!credits">
-          @if (credits && credits.length > 0) {
-            <h3 class="font-display font-bold">Credits</h3>
-            <div class="ng-star-inserted grid gap-2 md:grid-cols-3 lg:grid-cols-4">
-              @for (type of MapCreditType | enumValue; track $index) {
-                @if (creditMap[type]) {
-                  <div>
-                    <p class="font-sm mb-1 font-bold uppercase text-gray-200">
-                      {{ MapCreditNames.get(type) }}
-                    </p>
-                    @for (credit of creditMap[type]; track $index) {
-                      <a routerLink="/maps/{{ credit.map?.id }}">
-                        <p class="font-sm text-gray-200">
-                          {{ credit.map?.name }}
-                        </p>
-                      </a>
+        @if (isLocal) {
+          <div>
+            <button class="btn btn-blue w-fit" type="button" (click)="showSteamFriendsModal()">
+              Import Steam Friends
+              <m-icon icon="steam" pack="si" class="ml-2 shrink-0" />
+            </button>
+          </div>
+
+          <p-dialog
+            [mSpinner]="!steamFriends"
+            [(visible)]="importSteamFriendsModalVisible"
+            [modal]="true"
+            [style]="{ 'min-width': '25rem' }"
+            [draggable]="false"
+            [resizable]="false"
+          >
+            <ng-template pTemplate="header">
+              <div class="flex w-full flex-col gap-4">
+                <p class="text-lg">Import Steam friends to followers</p>
+                @if (steamFriends && steamFriends.length > 0) {
+                  <div class="flex justify-start gap-4">
+                    <button
+                      [disabled]="steamFriendsToAdd.size === steamFriends.length"
+                      class="btn btn-blue"
+                      type="button"
+                      (click)="selectAllSteamFriends()"
+                    >
+                      Select All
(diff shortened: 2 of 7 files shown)


apps/backend/src/app/modules/users/users.service.ts
```diff
@@ -747,5 +775,25 @@ export class UsersService {
...
+        if (!user) return null;
+
+        const isFollowing = await this.getFollower(userID, user.id);
+
+        if (isFollowing) return null;
+
+        return DtoFactory(UserDto, user);
+      })
+    ).then((users) => users.filter(Boolean));
+  }
```
tsa96, line 796: Shouldn't need to do multiple queries here, Prisma has an `in: ` property 
```ts
where: {
  steamID: {
    in: steamFriends.map(({ steamid }) => BigInt(steamid))
  }
}
```
that should correspond to a SQL `WHERE IN`.

Similar you don't want to do a `getFollower` for each friend since that'll be a separate query as well, should be able to do a similar Prisma query for that. In fact it's probably possible to do this whole thing in a Prisma single query though not off the top of my head sure how, so not essential. A case where I kinda wish we weren't using Prisma at all here since I'd rather just write this in regular SQL. But don't want to start integrating pure SQL until post 0.10.0... :(

apps/frontend/src/app/pages/profile/profile.component.html
```diff
@@ -154,29 +154,94 @@ <h5 class="m-0">
...
-                        <p class="font-sm text-gray-200">
-                          {{ credit.map?.name }}
-                        </p>
-                      </a>
+        @if (isLocal) {
+          <div>
+            <button class="btn btn-blue w-fit" type="button" (click)="showFriendsModal()">
+              Find Friends
+
+              <m-icon icon="steam" pack="si" class="ml-2 shrink-0" />
```
tsa96, line 162: I think `Import Steam Friends` is better here.

apps/frontend/src/app/pages/profile/profile.component.html
```diff
@@ -154,29 +154,94 @@ <h5 class="m-0">
...
+            [mSpinner]="!steamFriends"
+            [(visible)]="friendsModalVisible"
+            [modal]="true"
+            [style]="{ 'min-width': '25rem' }"
+            [draggable]="false"
+            [resizable]="false"
+          >
+            <ng-template pTemplate="header">
+              <div class="flex w-full flex-col gap-4">
+                <p class="text-lg">Add steam friends to followers</p>
```
tsa96, line 176: Capital "S" in Steam, also let's use "Import" instead of "Add"

apps/frontend/src/app/pages/profile/profile.component.html
```diff
@@ -154,29 +154,94 @@ <h5 class="m-0">
...
+            <button class="btn btn-blue w-fit" type="button" (click)="showFriendsModal()">
+              Find Friends
+
+              <m-icon icon="steam" pack="si" class="ml-2 shrink-0" />
+            </button>
+          </div>
+
+          <p-dialog
+            [mSpinner]="!steamFriends"
+            [(visible)]="friendsModalVisible"
```
tsa96, line 168: Nit: name is a bit ambiguous, I'd call it `importFriendsModalVisible`

apps/frontend/src/app/pages/profile/profile.component.ts
```diff
@@ -269,4 +287,52 @@ export class ProfileComponent implements OnInit {
...
+        });
+      },
+      error: (httpError: HttpErrorResponse) =>
+        this.messageService.add({
+          severity: 'error',
+          summary: 'Could not add friends',
+          detail: httpError.error.message
+        })
+    });
+  }
```
tsa96, line 337: Again with naming, I think it's okay to keep this code in this class but given how large the profile component may get we should use "steamFriends" not "friends" and similar

apps/frontend/src/app/pages/profile/profile.component.ts
```diff
@@ -269,4 +287,52 @@ export class ProfileComponent implements OnInit {
...
+      this.friendsToAdd.push(user.id);
+    }
+  }
+
+  selectAllFriends() {
+    this.friendsToAdd = this.steamFriends.map((user) => user.id);
+  }
+
+  deselectAllFriends() {
+    this.friendsToAdd = [];
```
tsa96, line 308: Bit cleaner to use a `Set` for `friendsToAdd` here instead of array.

tsa96 (changes requested):
Looking good! Don't have many issues, nice work with backend and E2E stuff.

pedroalves80:
All fixed @tsa96, u can review when u have a chance.

apps/backend/src/app/modules/users/users.service.ts
```diff
@@ -480,6 +480,34 @@ export class UsersService {
...
+
+        return this.db.follow.create({
+          data: { followeeID: userID, followedID: user.id }
+        });
+      })
+    );
+
+    return follows.map((follow) => DtoFactory(FollowDto, follow));
+  }
+
```
tsa96, line 510: Sorry I must've forgotten to flag this -- you have the same issue here as with my previous comment, doing individual queries for each follow.

The "already following" thing is probably because Prisma will throw if a Follow record already exists, but in that case you could first do a query to check where the user already follows any of the current, and throw if they do.

The self-following thing can done with just `if (userIDs.includes(userID)) throw new ... `

apps/backend/src/app/modules/users/users.service.ts
```diff
@@ -747,5 +775,38 @@ export class UsersService {
...
+      where: {
+        followeeID: userID,
+        followedID: { in: userIDs }
+      }
+    });
+    const followingSet = new Set(
+      followerRelations.map((rel) => rel.followedID)
+    );
+
+    const steamIDSet = new Set(steamIDs.map((id) => id));
```
tsa96, line 801: We don't need this code but just wanted to point out because it's cute: `steamIDs.map(id => id)` is the identity function, just copies the array.

apps/backend/src/app/modules/users/users.service.ts
```diff
@@ -747,5 +775,38 @@ export class UsersService {
...
+    const filteredUsers = users
+      .filter(
+        (user) => steamIDSet.has(user.steamID) && !followingSet.has(user.id)
+      )
+      .map((user) => DtoFactory(UserDto, user));
```
tsa96, line 806: I think this whole query can be just
```ts
    this.db.user.findMany({
      where: {
        steamID: { in: steamIDs },
        follows: { none: { followeeID: userID } }
      }
    });
```
though I'm half asleep atm so maybe that's wrong.

tsa96 (changes requested):

pedroalves80:
@tsa96 fixed all those issues. While I was at it I found that the followers service was being duplicated alot when going out and in of profile page, fixed that too.

tsa96 (approved):

apps/backend/src/app/modules/users/users.service.ts
```diff
@@ -480,6 +480,48 @@ export class UsersService {
...
+        followedID
+      }))
+    });
+
+    const createdFollows = await this.db.follow.findMany({
+      where: {
+        followeeID: userID,
+        followedID: { in: usersIDs }
+      }
+    });
```
tsa96, line 520: Sorry, last thing I promise. This prompted me to look up whether Prisma *finally* implemented a way to do it one query (though maybe it does multiple under the hood, you never know with Prisma). Turns out they have, since 5.14, so you should be able to use `createManyAndReturn` for this.

pedroalves80: No worries @tsa96! Fixed this issue too.

PR #1099, opened 2025-02-25
fix(shared): allow updating only one file from map version

### LINKED ISSUE #1080: Allow updating only map zones
I thought I did this, maybe backend allows, frontend form validation errors when just adding changelog + zones, needs a BSP to be valid.

pedroalves80:
Closes #1080 

Changed map version submission and added a validator in frontend;
Changed service to be prepared to not have a bsp in temp in backend;

image
image
image

### Checks

### DIFF
--- a/apps/backend-e2e/src/maps.e2e-spec.ts
+++ b/apps/backend-e2e/src/maps.e2e-spec.ts
@@ -2240,19 +2240,12 @@ describe('Maps', () => {
         });
       });
 
-      it('should 400 if BSP file is missing', async () => {
+      it('should succed if only the VMF file is provided', async () => {
         await req.postAttach({
           url: `maps/${map.id}`,
-          status: 400,
-          data: { changelog: 'help me' },
+          status: 201,
+          data: { changelog: 'who needs tests anyway' },
           files: [{ file: vmfBuffer, field: 'vmfs', fileName: 'surf_map.vmf' }],
-          token: u1Token
-        });
-
-        await req.post({
-          url: 'maps',
-          status: 400,
-          body: { changelog: 'im bored' },
           token: u1Token
         });
       });
@@ -2264,6 +2257,16 @@ describe('Maps', () => {
           url: `maps/${map.id}`,
           status: 201,
           data: { changelog: 'who needs tests anyway' },
+          token: u1Token
+        });
+      });
+
+      it('should succed if BSP file is missing', async () => {
+        await req.postAttach({
+          url: `maps/${map.id}`,
+          status: 201,
+          data: { changelog: 'who needs tests anyway' },
+          files: [{ file: vmfBuffer, field: 'vmfs', fileName: 'surf_map.vmf' }],
           token: u1Token
         });
       });
@@ -2282,6 +2285,15 @@ describe('Maps', () => {
               fileName: 'surf_map.vmf'
             }
           ],
+          token: u1Token
+        });
+      });
+
+      it('should 400 if neither VMF nor BSP nor zones are provided', async () => {
+        await req.postAttach({
+          url: `maps/${map.id}`,
+          status: 400,
+          data: { changelog: 'just hope it works' },
           token: u1Token
         });
       });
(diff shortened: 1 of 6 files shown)


apps/backend/src/app/modules/maps/maps.service.ts
```diff
@@ -539,6 +539,9 @@ export class MapsService {
 
     const bspFile = await this.getBspFromTemp(userID);
 
+    if (bspFile instanceof Error)
```
tsa96, line 542: This pattern of returning an `Error` is a bit unusual, I would just return `null`. Returning errors as regular objects isn't *necessarily* a bad thing but confused me a lot at first.

tsa96: Also while we're here let's give this function a clearer name, let's use `tryGetBspFromTempUploads`.

apps/backend/src/app/modules/maps/maps.service.ts
```diff
@@ -1891,13 +1907,13 @@ export class MapsService {
...
 
-  async getBspFromTemp(userID: number): Promise<File> {
+  async getBspFromTemp(userID: number): Promise<File | Error> {
     const userIDObjects = await this.fileStoreService.listFileKeys(
       `upload_tmp/${userID}`
     );
 
     if (userIDObjects.length === 0) {
-      throw new BadRequestException('Missing BSP file');
+      return new Error('Missing BSP file');
```
tsa96, line 1916: As above, let's return `null` here, not an error (which are typically reserved for `try/catch`)

apps/backend-e2e/src/maps.e2e-spec.ts
```diff
@@ -2240,23 +2240,6 @@ describe('Maps', () => {
         });
       });
 
-      it('should 400 if BSP file is missing', async () => {
```
tsa96, line 2243: We still want to require at least one of BSP/zones/VMF. So let's invert this test to check we allow missing BSP if there's no vmf/zones, then update service logic to throw if all three are missing, and test for that in here as well.

apps/frontend/src/app/pages/maps/map-edit/map-edit.component.html
```diff
@@ -31,7 +31,12 @@
         <span class="text-lg">Reset leaderboards</span>
         <input type="checkbox" class="checkbox" [formControl]="resetLbs" />
       </div>
-      <button type="submit" class="btn btn-green self-end" [disabled]="loading || versionForm.pristine || versionForm.invalid">
+
```
tsa96, line 34: Shouldn't need this, a form is invalid is it has errors.

apps/frontend/src/app/pages/maps/map-edit/map-edit.component.ts
```diff
@@ -184,24 +185,27 @@ export class MapEditComponent implements OnInit, ConfirmDeactivate {
...
+        [FileValidators.maxSize(MAX_VMF_SIZE), FileValidators.extension('vmf')],
+        [FileValidators.isValidVdf()]
+      ],
+      changelog: [
+        '',
+        [Validators.required, Validators.maxLength(MAX_CHANGELOG_LENGTH)]
+      ],
+      resetLbs: [false]
+    },
+    { validators: atLeastOneFileRequired(['bsp', 'zon']) }
```
tsa96, line 207: Should allow for VMFs as well, could imagine someone only uploading a BSP and not VMF, then deciding to the VMF after.

apps/frontend/src/app/validators/map-edit.validator.ts
```diff
@@ -0,0 +1,10 @@
+import { FormGroup } from '@angular/forms';
+
+export function atLeastOneFileRequired(fieldNames: string[]) {
```
tsa96, line 3: Validator is good but isn't specific to files. Also may as well make it more flexible, use a `filter(...).length` and have it be `atLeastNRequired(numRequired: number, fieldNames: string[])`

pedroalves80:
All done @tsa96. LMK if it needs any more fixing

PR #1122, opened 2025-03-08
fix(back-e2e): disconnect Prisma clients in teardownE2ETestEnvironment

tsa96:
Prisma started complaing about having too many clients open when trying to run all E2E tests local after recent dependency upgrades. Turns out I'm a complete moron and was never disconnecting the Prisma clients used by each test.

### DIFF
--- a/apps/backend-e2e/src/activities.e2e-spec.ts
+++ b/apps/backend-e2e/src/activities.e2e-spec.ts
@@ -22,7 +22,7 @@ describe('Activities', () => {
     db = env.db;
   });
 
-  afterAll(() => teardownE2ETestEnvironment(app));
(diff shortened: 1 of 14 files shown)


GordiNoki:
Some useless information: In my case every opened prisma client is opening 2 postgres connections. So, I noticed that by default - yea all the clients disconnect only after the whole suite ran, which does mean we need to close them in between runs, but you currently only close ones that are used to modify database in tests, there is another client in nest app, which stays connected even after running `app.close()`. With your current changes you do cut amount of stale connections by two, but I think you should also run `app.get(EXTENDED_PRISMA_SERVICE).$disconnect()` there

GordiNoki (approved):

Issue #855, opened 2023-10-15
Enable exactOptionalPropertyTypes in TypeScript


tsa96:
This is a sensible flag that helps TS distinguish between fields that can be omitted and fields that can take a value of `undefined`. Enabling it stops you being able to set e.g. `k?: string` to `undefined`, and makes TypeScript's type inference when using index signatures a lot stronger - see https://github.com/microsoft/TypeScript/pull/43947 

For us, it allows us to make nested interfaces containing optional values using `extends JsonObject` from type-fest, currently it just spews out errors about not being able to assign to some (e.g. `string`) index type, see https://github.com/sindresorhus/type-fest/issues/471#issuecomment-1284987356 for details.

TypeScript unfortunately doesn't have it on by default or in strict mode, but I'd definitely like to support it. I would add it right now, but it requires `--strictNullChecks`. So, once we finish https://github.com/momentum-mod/website/issues/792, this is a freebie.

I have to do a bunch of ugly `as unknown as MapZones` casts at the moment because of it, leaving comments everywhere that I am referencing this issue number. Once we tackle this, search for every instance and remove.

tsa96:
>> I have to do a bunch of ugly as unknown as MapZones casts at the moment because of it, leaving comments everywhere that I am referencing this issue number. Once we tackle this, search for every instance and remove.

I was getting this wrong -- the issue is using interfaces over types, where using types would given us [implicit index signatures](https://www.totaltypescript.com/type-vs-interface-which-should-you-use#index-signatures-in-types-vs-interfaces) that satisfy Prisma's JSON types. Not 100% sure on best approach atm, tempted to convert everything to types but if we're getting rid of class-transformer DTO classes anyway then adding index signatures to every interface doesn't seem that bad. Worth thinking about this more when deciding on future approaches to DTOs and validation.

PR #1148, opened 2025-04-06
Fix wonky credits

### LINKED ISSUE #1147: Can't update map credits if only placeholder suggestions as authors
Check in `updateCredits` in backend currently requires a non-placeholder-suggestion with an author credit, check needs to allow for placeholder suggestions.

Frontend has then needs tweaking to ensure the requests get sent in the right order.

tsa96:
Closes #1147

### Checks

### DIFF
--- a/apps/backend/src/app/modules/maps/map-credits.service.ts
+++ b/apps/backend/src/app/modules/maps/map-credits.service.ts
@@ -13,7 +13,8 @@ import {
   MapStatuses,
   CombinedRoles,
   MapCreditsGetExpand,
-  MapCreditType
+  MapCreditType,
+  MapSubmissionPlaceholder
 } from '@momentum/constants';
 import { findWithIndex } from '@momentum/util-fn';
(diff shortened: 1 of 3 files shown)


apps/backend/src/app/modules/maps/map-credits.service.ts
```diff
@@ -96,13 +94,24 @@ export class MapCreditsService {
...
     if (!map) {
       throw new NotFoundException('Map not found');
     }
 
+    if (
+      !body.some((credit) => credit.type === MapCreditType.AUTHOR) &&
+      !(
+        MapStatuses.IN_SUBMISSION.includes(map.status) &&
+        (
+          map.submission.placeholders as unknown as MapSubmissionPlaceholder[]
```
GordiNoki, line 109: nit: I'd like there to be a `// TODO: #855` comment since it would be easier to find all of the places we do unknown cast because of exactOptionalPropertyTypes

GordiNoki (approved):

PR #1149, opened 2025-04-09
Make map lists generate on schedule, instead of map updates

### LINKED ISSUE #1078: Throttle map list generation
Static map list generation is quite slow, especially for large maps, we limit various aspects of map submission but someone could potentially force lots of list generations by spamming status changes.

We could fight this on a very granular level, but I think it's probably easiest to just limit how frequently lists are generated. Have a constant minimum time between list regens, then if a regen is requested below that time, schedule for after that time. Maybe best  to do in RxJS.

GordiNoki:
Closes #1078

### Checks

### DIFF
--- a/apps/backend-e2e/src/maps.e2e-spec.ts
+++ b/apps/backend-e2e/src/maps.e2e-spec.ts
@@ -2,6 +2,7 @@
 
 import { Config } from '../../backend/src/app/config';
 import { MapDto } from '../../backend/src/app/dto';
+import { MAPLIST_UPDATE_JOB_NAME } from '../../backend/src/app/modules/maps/map-list.service';
 import { readFileSync } from 'node:fs';
 import path from 'node:path';
 import crypto from 'node:crypto';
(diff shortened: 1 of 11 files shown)


apps/backend-e2e/src/maps.e2e-spec.ts
```diff
@@ -66,6 +68,10 @@ describe('Maps', () => {
     req = env.req;
     db = env.db;
     fileStore = env.fileStore;
+    const mapListService = env.app.get(MapListService);
+    checkScheduledUpdates = (mapListService as any)[
+      'checkScheduledUpdates'
+    ].bind(mapListService);
```
tsa96, line 74: Shame we can't use faketimers for this... they break Postgres and MinIO in E2E tests unfortunately.

tsa96 (approved):
Looks good, nice work.

apps/backend-e2e/src/maps.e2e-spec.ts
GordiNoki: Well, if we wanna be a bit more fancy, there is a way to get a cron job from scheduler registry and run it manually

tsa96: If that's possible, let's go for it. Good to avoid such an ugly approach especially if we end up doing more scheduler stuff in the future.

tsa96 (approved):

PR #1154, opened 2025-04-19
Make auth call Steam's GetPlayerSummaries much less

GordiNoki:
Before we were calling steam for user data every time user tries to authenticate to update out saved data, but it doesn't make much sense. So now we would call steam only for user account creation

### Checks


tsa96 (approved):
So much deleted, nice work!!

Gocnak:
What if someone updates their name on steam? Do we store their username still? If we don't store it then this is fine, otherwise we need to be able to update their name somehow

PR #1164, opened 2025-05-13
fix(front): update review list in the case we comment on a deleted review

N333kk:
Added an adecuate message when the review has been deleted, also made it so the review list updates in case the error is triggered.

### DIFF
--- a/apps/backend-e2e/src/map-review.e2e-spec.ts
+++ b/apps/backend-e2e/src/map-review.e2e-spec.ts
@@ -489,6 +489,18 @@ describe('Map Reviews', () => {
 
       afterAll(() => db.cleanup('mMap', 'user'));
 
+      it('should 400 if commenting on non-existing review', () =>
+        req.post({
+          url: `map-review/${NULL_ID}/comments`,
+          status: 400,
+          body: {
+            data: {
+              text: 'Not disgusting!!'
+            }
+          },
+          token: token
+        }));
+
       it('should 503 if the killswitch is true', async () => {
         await req.patch({
           url: 'admin/killswitch',
--- a/apps/backend/src/app/modules/map-review/map-review-comment.service.ts
+++ b/apps/backend/src/app/modules/map-review/map-review-comment.service.ts
@@ -137,6 +137,12 @@ export class MapReviewCommentService {
       include: { mmap: true }
     });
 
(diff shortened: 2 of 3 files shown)


apps/backend-e2e/src/map-review.e2e-spec.ts
```diff
@@ -489,6 +489,32 @@ describe('Map Reviews', () => {
...
+          url: `map-review/${tempReview.id}/comments`,
+          status: 400,
+          body: {
+            data: {
+              text: 'Not disgusting!!'
+            }
+          },
+          token: token
+        });
+        expect(response.body.message).toMatch(
```
tsa96, line 513: Don't need to do an exact string check here, bit brittle and important thing is just that we return 400.

N333kk: Just changed it to expect a 400 status code.

apps/backend-e2e/src/map-review.e2e-spec.ts
```diff
@@ -489,6 +489,32 @@ describe('Map Reviews', () => {
 
       afterAll(() => db.cleanup('mMap', 'user'));
 
+      it('should return apropiate error if review has been deleted', async () => {
```
GordiNoki, line 492: Just do this:
```ts
      it('should 400 if commenting on non-existing review', () =>
        req.post({
          url: `map-review/${NULL_ID}/comments`,
          status: 400,
          body: {
            data: {
              text: 'Not disgusting!!'
            }
          },
          token: token
        }));
```

All the request utilities have status checks, so if you set the `status` field here to 400 it will check that. Also you don't need to delete a review to test this. Just providing a non-existing id should be fine

GordiNoki (changes requested):
Don't make a new commit to fix stuff asked in a review. You should fixup your changes to an original commit

apps/backend-e2e/src/map-review.e2e-spec.ts
N333kk: Changed this right now, sorry about the fixup thing.

GordiNoki (approved):

tsa96 (approved):

tsa96:
Looks good, thanks for this!

PR #1166, opened 2025-05-19
chore(back): Remove unused const and check

Bucketheaddd:
Closes #

Removing definition and usage of `MAX_ZONES_ALL_TRACKS` as it is no longer needed.

### DIFF
--- a/libs/formats/zone/src/zone-validator.spec.ts
+++ b/libs/formats/zone/src/zone-validator.spec.ts
@@ -431,7 +431,7 @@ describe('validateZoneFile', () => {
     });
 
     // Actually impossible to reach this, since only main tracks can have
-    // segments, and the MAX_ZONES_ALL_TRACKS check will catch it first.
+    // segments, and the MAX_REGIONS check will catch it first.
     it.skip('should throw if a track has too many segments', () => {});
 
     it('should throw if a segment has no checkpoints', () => {
--- a/libs/formats/zone/src/zone-validator.ts
+++ b/libs/formats/zone/src/zone-validator.ts
@@ -28,12 +28,6 @@ export const MAX_REGIONS = 250;
 // Includes the segment start zone
 export const MAX_SEGMENT_CHECKPOINTS = 100;
 
-// From C++:
-// One segment start for max Main segments plus Main end zone plus start and end for each stage track.
-// The limit could be increased from here (barring region entity overload) but this is the bare minimum to allow for MAX_TRACK_SEGMENTS.
-export const MAX_ZONES_ALL_TRACKS =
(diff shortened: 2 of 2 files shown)


tsa96 (changes requested):
Thanks for the PR.

To get unit tests to pass you'll need to remove the failing tests related to MAX_ZONES_ALL_TRACKS, and to make this equivalent C++, the incrementing `totalZones` variable in C++ needs to be replaced by `totalRegions`, which increments for every region the validator iterates over, then have an equivalent check where the `totalZones` one was, i.e. 
```ts
  if (totalRegions > MAX_REGIONS)
    throw new ZoneValidationError('Too many regions in total');
```
then just do a unit test in the same fashion of the totalZones one.

Sorry if rio's issue was a bit unclear, the corresponding C++ validation logic unfortunately has to be closed-source.

Oh also, Prettier is complaining about autoformatting for some reason, can just run `nx format` to fix everything. And please fixup/amend any changes into your original commit! Thanks :)

Bucketheaddd:
> To get unit tests to pass you'll need to remove the failing tests related to MAX_ZONES_ALL_TRACKS, and to make this equivalent C++, the incrementing `totalZones` variable in C++ needs to be replaced by `totalRegions`, which increments for every region the validator iterates over, then have an equivalent check where the `totalZones` one was, i.e.

Is this check intended to limit the number of regions per zone or the number of regions on all tracks (i.e regions per zone per segment)? Making the change to the test as I understand it is actually failing a test when the `MAX_TRACK_SEGMENTS` check happens. If 100 segments are allowed, then each one would be limited to 2-3 regions each. 

So I'm assuming the `totalRegions` needs to incriment per zone?

tsa96:
Oh heads up, that failing E2E test is just a flaky test we haven't fixed yet, can ignore. The linter thing needs addressing though, `nx lint formats-zone` should do it (see https://github.com/momentum-mod/website/wiki/Setup#other-commands for commands)

Bucketheaddd:
I was struggling to get the tests to pass locally and I think it is because counting the regions in the `validateZone` function was causing overcounts. I added a manual count before the other validations and updated the tests. I ran the linter locally and it passed. Did not make any changes to address the E2E per your last comment. 

I hope everything passes this time. Happy to be contributing.

jason-e:
> BTW @jason-e I'm struggling to understand the C++, do a main track and stages share the same regions, or are they distinct?

Sorry for missing this. They do share regions. The idea with that and with limiting regions overall (vs limiting zones or something else) is because each region is an entity, so we want to keep the number of entities created sane. Networking is not really a concern since only region entities in the current PVS are networked/recorded.

Also it does look like the latest iteration here is not accounting for global regions.

PR #1177, opened 2025-06-05
Only update a replay file if run is a PB

GordiNoki:
Fixes an error where some replays would return 404. Previously it was deleting current PB run every time a new run was submitted, and saving a new one instead. Since we should only store PB runs, now it will call a function to save a new replay and delete a previous one only if a new run is a PB.

### Checks

### DIFF
--- a/apps/backend/src/app/modules/session/run/run-session.service.ts
+++ b/apps/backend/src/app/modules/session/run/run-session.service.ts
@@ -13,7 +13,6 @@ import {
   TrackType,
   XpGain
 } from '@momentum/constants';
-import { parallel } from '@momentum/util-fn';
 import { FileStoreService } from '../../filestore/file-store.service';
 import { XpSystemsService } from '../../xp-systems/xp-systems.service';
 import {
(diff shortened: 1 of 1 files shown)


apps/backend/src/app/modules/session/run/run-session.service.ts
```diff
@@ -251,12 +251,25 @@ export class RunSessionService {
 
     const replayHash = FileStoreService.getHashForBuffer(replayBuffer);
 
+    const tasks: [
```
tsa96, line 254: Types are a little ugly, hurts code readability. Does a `satisfies PromiseLike<any>[]` work instead, then I let ts infer original type?

apps/backend/src/app/modules/session/run/run-session.service.ts
```diff
@@ -251,12 +251,25 @@ export class RunSessionService {
...
+        )
+      );
+
     // We have two quite expensive, independent operations here, including a
     // file store. So we may as well run in parallel and await them both.
-    const [{ run, xpGain, isWR }] = await parallel(
-      this.updateLeaderboards(tx, submittedRun, isPB, existingRun, replayHash),
-      this.updateReplayFiles(replayBuffer, replayHash, existingRun?.replayHash)
-    );
+    const [{ run, xpGain, isWR }] = await parallel(...tasks);
```
tsa96, line 272: Honestly, my parallel thing is just dumb and should probably phase out - you can just do Promise.all(tasks)

tsa96 (changes requested):

apps/backend/src/app/modules/session/run/run-session.service.ts
GordiNoki: I couldn't figure it out with satisfies. It was giving me an error when I tried to push another promise to an array with just the satisfies keyword, so I decided to ditch the array since it makes stuff to complex.

tsa96 (approved):

PR #1179, opened 2025-06-10
409 on duplicate BSP already used by other map

### LINKED ISSUE #245: Hash the map on the backend and compare against other map hashes to prevent duplicate file uploading

AshenCinders:
closes #245 

revival of #1082 

POST/maps and POST/maps/MAPID now throws 409 if another map uses the exact same BSP version.

### Checks


AshenCinders:
question: is it preferred that I split up e2e test in their own commits (in the future) since it would use backend-e2e scope rather than backend?

apps/backend-e2e/src/maps.e2e-spec.ts
```diff
@@ -1872,42 +1914,41 @@ describe('Maps', () => {
...
-          {
-            name: 'surf_map', // This is actually RJ now. deal with it lol
-            submitter: { connect: { id: u1.id } },
-            status: MapStatus.PRIVATE_TESTING,
-            versions: {
-              create: {
-                zones: ZonesStubString,
-                versionNum: 1,
-                bspHash: createSha1Hash(Buffer.from('hash browns')),
-                submitter: { connect: { id: u1.id } }
```
AshenCinders, line 1885: test below needs to have the same bspHash as bspHash variable above, alternatively I could copy the definition here into my test but that seems ugly.
No tests fail from the change to I've done this instead
changed to `createSha1Hash(bspBuffer)`

apps/backend-e2e/src/maps.e2e-spec.ts
```diff
@@ -729,23 +729,27 @@ describe('Maps', () => {
     });
 
     describe('POST', () => {
-      let user,
-        token,
+      let user: User,
+        token: string,
         createMapObject,
```
AshenCinders, line 734: the any-s annoyed me so added types, couldn't figure out the more complex ones though

apps/backend-e2e/src/maps.e2e-spec.ts
tsa96: Yeah, this is fine!

tsa96 (approved):

tsa96:
> question: is it preferred that I split up e2e test in their own commits (in the future) since it would use backend-e2e scope rather than backend?

I wouldn't really worry about it unless the e2e changes are enormous, what you did is fine!

GordiNoki (approved):

PR #1182, opened 2025-06-11
Fix not being able to set verified role

### LINKED ISSUE #1139: Impossible to verify users due to alias collision
If you try to set a user to verified, the DB call that checks whether a verified user exists with that alias *includes the target user*, so fails every time.

To replicate, start a seeded DB, give yourself admin, try to add Verified role to any user.

Should be easy to fix, but make sure to include E2E tests that verify this is fixed.

Not needed for 0.10.0 but fairly high priority since Verified role is completely unusable without this.

AshenCinders:
closes #1139 

now allows admins to set user as verified role

also fixed a hidden bug where a user would still get assigned verified role even if there was an alias collision, since actual alias collision check was done after admin-function set new role in db.

I have checked the equivalent that the tests do directly in the browser, and the fail notifications come up in the correct priority.

### Checks


apps/backend/src/app/modules/admin/admin.service.ts
```diff
@@ -237,6 +238,23 @@ export class AdminService {
...
+        if (
+          sameNameMatches.some((user) =>
+            Bitflags.has(user.roles, Role.VERIFIED)
+          )
+        ) {
+          throw new ConflictException(
+            'Cannot give user verified role since alias is already in use by another verified user'
+          );
+        }
+      }
```
AshenCinders, line 256: the added logic is effectively the same as is done in user function.
Though this one throws for trying to change role, while in user function throws for trying to change alias.
I thought it would be worse to try to abstract out these as a function, since I'd need a string param for exception message as well.

AshenCinders: on second thought I could pass the whole exception in an arrow func, though probably still not better than current solution?

PR #1185, opened 2025-06-13
chore(deps): bump dependencies to fix audit vulnerabilities

GordiNoki:
Makes some funny numbers go smaller
`27 vulnerabilities (4 low, 14 moderate, 3 high, 6 critical)`
↓
`23 vulnerabilities (4 low, 19 moderate)`


package.json
```diff
@@ -89,10 +89,10 @@
...
     "@eslint/js": "~9.19.0",
     "@faker-js/faker": "9.4.0",
-    "@nestjs/cli": "11.0.2",
-    "@nestjs/schematics": "11.0.0",
-    "@nestjs/testing": "11.0.5",
-    "@nx/angular": "20.3.3",
+    "@nestjs/cli": "11.0.7",
+    "@nestjs/schematics": "11.0.5",
+    "@nestjs/testing": "11.1.3",
+    "@nx/angular": "21.2.0",
```
tsa96, line 95: Have you checked whether we need Nx migrations for this? Unfortunately since Nx keeps all its packages on the same versions, it's generally best to let Nx bump and migrate everything at once, though that can cause dependency issues with other packages. Going to try a migration now and see if anything breaks.

tsa96:
Since we're so close to full prod we should probably start relying on dependabot for this.

It's currently configured to only create PRs for security updates, so less annoying than it used to be, but it still creates security updates for obscure things in dev tools, which don't tend to matter that much. I'm going to try reconfigure it to separate PRs for prod and dev node deps, then in future, we should immediately resolve any production vulnerabilities, doing a full Nx upgrade if it affects a Nx-managed dependency, rather than use dependabot.

PR #1196, opened 2025-06-15
fix(front): remove animation on profile xp bar

### LINKED ISSUE #1169: Player XP bar on profile bar animates when viewing different profiles
Angular is reusing this component (good) but produces this animation (bad):

https://github.com/user-attachments/assets/d78fac3e-9737-4c73-8dae-d4aaa2af2cad

Doesn't seem to be any reason to animate this in the first place, user XP doesn't change based on actions on the site, just remove the animation.

RogiJAG:
Closes #1169 

Disables the value changed animation which was previously triggered when switching between two profiles with different xp values. 

### Screenshots

No visual changes to xp bar. See #1169 for original behavior.

image

#### Comments
NG doesn't support disabling this natively - opted for a css override. Imo it's small enough to be inline, but if it needs to be moved let me know.

### DIFF
--- a/apps/frontend/src/app/theme/primeng-preset.ts
+++ b/apps/frontend/src/app/theme/primeng-preset.ts
@@ -55,7 +55,8 @@ export const MomentumPreset = definePreset(Material, {
       value: {
         background:
           'linear-gradient(90deg, rgb(var(--blue-500)), rgb(var(--blue-300)))'
-      }
+      },
+      css: () => '.p-progressbar-value { transition: none !important; }'
     },
     dialog: {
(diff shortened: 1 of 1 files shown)


GordiNoki (changes requested):
Thanks for a contribution. I really don't like the inline style block. You could put it in the styles attribute in component .ts file, but I think those styles should actually be in a primeng-preset.ts file. I don't think primeng themes support disabling transitions, so you can just apply a custom css style to a progress bar component. There are examples of how to do it in this file, so you can just do the same.

GordiNoki (approved):
Good job

PR #1218, opened 2025-07-12
Map browser filters persistence

### LINKED ISSUE #1213: Make map browser filters save between navigation
When clicking around in map browsers; every time you click on a map or load a different page and go back the filters reset.
A common use case for when you would want the filters to stay, would for example be when sorting for most favorited, and looking at each map in order. (Or uncompleted maps, etc.)
Currently you would have to reselect the filters each time, or open maps in new tabs.

This might be quite ugly to implement though.

I think it would be preferred to have each map browser's filters save independently, and that all filters reset on F5.
So selected filters only stay for the current session.

AshenCinders:
closes #1213 

For each map browser; now saves filters to sessionStorage, has clear filters button
Also changed m-dropdown to allow different widths, as it was ~hardcoded before.
Also changed number of inputs in multiselects for when to reduce to "x items selected" to 1 to stop wrapping of other elements if 3 long options were chosen (default amount to reduce is 4).

I decided to stick with sessionStorage to allow multiple tabs with different filters to be allowed,
and when closing a tab it seems reasonable to reset the filters as well (which is what sessionStorage does).
Wasn't able to get filters to reset on refresh as well, without some ugly changes, so decided to skip (as per discord comments).
Though, if we want to change to localStorage, its a 2-line change.

Regarding the p-multiselect change; if I were to make an option/variant of the homecooked m-dropdown for multi-selects, I think it would make sense to default the write-out limit to 1 as well, if y'all agree.

I tried making the clear-filters button same size as nearby boxes, but looked bad, so matched the other buttons instead.

Note that the main map browser dropdown gets extra chonky to keep the nice width alignment with the above bit.

vivaldi_F6J7wgQcYE
96-ab52-019549726f51" />
vivaldi_luwoVmep8z
vivaldi_JNpC5LI5ru

### Checks


AshenCinders:
Regarding documentation on stuff like the persistent form func.
The documentation in the codebase is very mixed in styles: none, jsdoc, `/** **/` without @ s, or just a // here and there.
Which style is preferred?

apps/frontend/src/app/pages/maps/browsers/map-browser.component.html
```diff
@@ -74,7 +74,7 @@
           formControlName="pb"
         />
       }
-      <m-dropdown [entries]="MapSortOptions" [entryNameFn]="MapSortNameFn" formControlName="sortType" />
+      <m-dropdown [entries]="MapSortOptions" [entryNameFn]="MapSortNameFn" formControlName="sortType" [ngStyle]="{ width: '18.5rem' }" />
```
tsa96, line 80: This can be a tailwind arbitrary value i.e. `w-[18.5rem]`.

apps/frontend/src/app/pages/maps/browsers/admin-maps-browser.component.html
```diff
@@ -13,6 +13,8 @@
       optionValue="type"
       [showHeader]="false"
       appendTo="body"
+      [maxSelectedLabels]="1"
+      class="w-48"
```
tsa96, line 17: Can't we at least allow two items here? If we're wrapping at that, we should maybe restyle outer stuff so wrapping behaves better.

apps/frontend/src/app/pages/maps/browsers/map-browser.component.ts
```diff
@@ -190,6 +209,21 @@ export class MapBrowserComponent implements OnInit {
       });
   }
 
+  ngOnDestroy() {
+    this.persistentFiltersSubscription.unsubscribe();
+  }
```
tsa96, line 214: Thinking about this more, the cleanup in destructor shouldn't be necessary: this subscription won't emit again, since it's derived from the form, which is destroy when this component is destroy. So you can remove this and `persistentFiltersSubscription`. Sorry for leading you astray!

apps/frontend/src/app/pages/maps/browsers/map-browser.component.ts
```diff
@@ -96,7 +105,17 @@ export class MapBrowserComponent implements OnInit {
   protected readonly initialItems = 16;
   protected readonly itemsPerLoad = 8;
 
+  private persistentFiltersSubscription: Subscription;
+  private readonly persistentFiltersKey = this.constructor.name + '_FILTERS';
```
tsa96, line 109: Does this need to be a field? Only being used in one place, I'd just inline.

tsa96:
> Regarding documentation on stuff like the persistent form func. The documentation in the codebase is very mixed in styles: none, jsdoc, `/** **/` without @ s, or just a // here and there. Which style is preferred?

JSDoc is best for any reusable function that's getting used in a bunch of places and it's helpful for a dev to read docs for. We don't use `@s` for anything that TypeScript handles like `@param`, `@return` etc.

We also don't tend to do docs for functions that are totally self-explanatory given their name and signature. Obviously that's a bit subjective, but at least avoid stuff like ```
```ts
/**
  * Returns Foo
  * @returns {Foo}
  */
export function getFoo(): Foo {
  return this.foo;
}
```

tsa96 (changes requested):

apps/frontend/src/app/pages/maps/browsers/map-browser.component.ts
tsa96: Maybe I'm confusing JS and C# Observables actually, I think RxJS observables are Observables are more annoying to dispose. [This seems like a decent reference.](https://stackoverflow.com/a/51732897). My Angular skills are getting rusty!

So for form observables I think you *do* need to unsubscribe (annoying), but we use a slightly cleaner `takeUntilDestroyed` pattern most of the time. This class already injects `DestroyRef`, let's pass that to `setupPersistentForm` and change it to
```ts
export function setupPersistentForm(
  form: AbstractControl,
  key: string,
  destroyRef: DestroyRef
): Subscription {
  const valueString = sessionStorage.getItem(key);

  if (valueString) {
    form.setValue(JSON.parse(valueString));
  }

  return form.valueChanges
    .pipe(takeUntilDestroyed(destroyRef))
    .subscribe(() => sessionStorage.setItem(key, JSON.stringify(form.getRawValue()))
  );
}
```

apps/frontend/src/app/pages/maps/browsers/admin-maps-browser.component.html
AshenCinders: 2 just about works for now, so I've changed to that.
I'll have to restyle the wrapper divs when I add more filters in the future

PR #1224, opened 2025-07-14
Fix/double submit again

### LINKED ISSUE #1129: Double Clicking "Post" when leaving a review uploads it twice
1. Go to a beta map page
2. Write a review
3. Double click post
4. It gets posted twice

**Expected behavior** 
Review should only be posted once

**Desktop:**

- OS: Linux
- Browser: Firefox

AshenCinders:
fixes #1129 again... sorry

Playing around with settimeout I was able to for example post a review twice, even with the `loading` check.
The `tap` function gets called before the code in `next/error` meaning a user could press post again just after loading was set to false in tap.
I've fixed the cases where I could actually prompt a double submit/delete/etc, but there are a few other functions left that still set a bool to false in tap. Not sure If we want to change these, as well.

Also fixed some tooltips not showing cause directive was missing.

Regarding setting the disabled attribute on the buttons (which I also did in 2 previous PRs), it's not necessary.
But, it will gray out the button to give feedback to the user that it is currently not clickable.
If you want that behaviour, I can change those buttons so that they gray out when loading. But it's up to you.
Currently, some of them do, some of them don't.

### Checks

### DIFF
--- a/apps/frontend/src/app/components/map-review/map-review-form.component.ts
+++ b/apps/frontend/src/app/components/map-review/map-review-form.component.ts
@@ -17,7 +17,7 @@ import {
   ReactiveFormsModule,
   Validators
 } from '@angular/forms';
-import { take, tap } from 'rxjs/operators';
+import { finalize, take } from 'rxjs/operators';
 import { FileValidators, suggestionsValidator } from '../../validators';
 import { MapReviewSuggestionsFormComponent } from './map-review-suggestions-form.component';
 import { SuggestionType } from '@momentum/formats/zone';
@@ -112,11 +112,11 @@ export class MapReviewFormComponent {
   protected loading = false;
 
   submit() {
-    if (this.form.invalid || this.form.pristine) return;
+    if (this.loading || this.form.invalid || this.form.pristine) return;
     this.loading = true;
+
     const suggestions =
       this.suggestions.value?.length > 0 ? this.suggestions.value : undefined;
-
     // Ignore any empty suggestions
     if (suggestions) {
       suggestions.forEach(({ tier, gameplayRating, tags }, i) => {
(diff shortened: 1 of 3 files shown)


AshenCinders:
a ffs why would it make a merge commit

apps/frontend/src/app/components/map-review/map-review-form.component.ts
```diff
@@ -141,39 +141,36 @@ export class MapReviewFormComponent {
...
+      next: () => {
+        this.reviewPosted.next();
+        this.messageService.add({
+          severity: 'success',
+          summary: `${this.editing ? 'Updated' : 'Posted'} map review for ${
+            this.map.name
+          }`
+        });
+        this.form.reset();
+        this.loading = false;
```
tsa96, line 154: Since we have a `take(1)` here the observable will complete after a single `next` or `error` call, so rather than do `this.loading: false` in both `next` and `error` callbacks, I'd do `complete: () => {this.loading = false;}`, bit easier to follow. Applies to other change as well.

tsa96 (changes requested):

apps/frontend/src/app/components/map-review/map-review-form.component.ts
AshenCinders: According to docs:
```
(property) complete?: () => void
A callback function that gets called by the producer if and when it has no more values to provide (by calling next callback function). This means that no error has happened. This callback can't be called more than one time, it can't be called if the error callback function have been called previously, nor it can't be called if the consumer has unsubscribed.
```

Though it seems I should be using `finalize` instead.
The change will end up being tap swapped for finalize.

tsa96 (approved):

PR #1227, opened 2025-07-15
MapInfo.approvedDate is a DateString, not Date

tsa96:
Just a tiny thing I noticed. We don't automatically parse ISO date strings on front or backend, they're just regular strings. Need to parse them manually to use as dates.


GordiNoki:
Check Type Tests. I did this because prisma generated client uses `Date | null` type for nullable dates instead of DateString

tsa96:
> Check Type Tests. I did this because prisma generated client uses `Date | null` type for nullable dates instead of DateString

Oh okay, `TransformPrismaType` needed updating to support `Date | null`.

GordiNoki (approved):

PR #1240, opened 2025-07-21
feat(front): display map submission version dates

AshenCinders:
Now displays (local) date for each map version in map submission page.

vivaldi_BikruEgQVV

### Checks

### DIFF
--- a/apps/frontend/src/app/pages/maps/map-info/map-submission/map-submission.component.html
+++ b/apps/frontend/src/app/pages/maps/map-info/map-submission/map-submission.component.html
@@ -52,7 +52,10 @@
   <legend class="text-shadow-strong font-display text-4xl font-bold text-gray-50">Versions</legend>
 
   @for (version of versions.slice(0, visibleVersions); track version.versionNum) {
-    <p class="mt-1 text-sm font-medium uppercase text-gray-200">Version {{ version.versionNum }}</p>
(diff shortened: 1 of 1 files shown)


AshenCinders:
@tsa96 the timeline already has date shown for each thing. Did you want there to be more specificity and non-local timestamp on hover?
Not sure I understood what you meant.

AshenCinders:
also do you have comments on:

1. "Don't know if there is a way to save the currently used dates (in prod db), as they are json."

2. "I've kept the same onDelete: cascade for users as other tables use, but I'm wondering if it would be better to use something like setNull instead?"

tsa96:
Oh sorry, missed those comments:
> Don't know if there is a way to save the currently used dates (in prod db), as they are json.
We could save them but patching the new table with that data wouldn't work very well since we haven't been storing the userIDs. Not super important information, fine if the existing stuff is missing in the new table.

> I've kept the same onDelete: cascade for users as other tables use, but I'm wondering if it would be better to use something like setNull instead?

Yeah use SetNull. MapSubmissionVersion should use SetNull for user deletion as well now I think about it.

AshenCinders:
@tsa96  changed user in MapSubmissionDate to be nullable.
I'm leaving the other one for the strict null issue

tsa96 (approved):

PR #1259, opened 2025-08-06
Move placeholder creation in map submission inside search window

### LINKED ISSUE #1230: Move create placeholder inside user search results in Map submission/Map edit forms
When adding credits to a map, and the user which is searched for does not have a MMod account, it will show empty results.
The next action then will always be to copy/re-write the name in the placeholder field and click Add.
The placeholder field should only be used when a MMod account does not exist.

If the Add button (and tooltip icon) is instead moved inside the results dropdown, the same input field can be used for the placeholder (so the submitter does not have to input username again).
And the Add placeholder button will only display when there is no existing MMod account, which will force the desired usage for it.

AshenCinders:
Closes #1230 

The placeholder validators didn't work before to begin with, can't figure out why.
~~For now, there is no feedback when trying to add a placeholder with length longer than max allowed.
(There wasn't any before for the old placeholder input either.)~~
Also displays error if max alias length exceeded.

### Checks

### DIFF
--- a/apps/frontend/src/app/pages/maps/map-info/map-submission/map-submission.component.html
+++ b/apps/frontend/src/app/pages/maps/map-info/map-submission/map-submission.component.html
@@ -14,10 +14,13 @@
   <p class="mt-4 text-sm font-medium uppercase text-gray-200">Timeline</p>
   <ul class="flex flex-wrap gap-x-8">
     @for (date of map.submission.dates; track $index) {
-      <li [mTooltip]="'Authored by ' + (date.user ? date.user.alias : 'N/A')">
+      <li [mTooltip]="timelineTooltip">
         <p class="font-medium">{{ MapStatusName.get(date.status) }}</p>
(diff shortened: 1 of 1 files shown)


tsa96 (changes requested):
Discussed when I was streaming, let's display the button unconditionally the user bit, on a separate row below everything else, rather than integrate with the "no users found" bit

AshenCinders:
The width for the whole popover has a static width that is relied on by other users of the component. So if I were to put the placeholder button below no-results-found text it looks like ass :D.

Now also displays on results found.

I've tried to minimise the styling changes needed for future use, so there is an extra div wrapper that the credits section uses.

vivaldi_8lcPIA2n5w
vivaldi_4F9OmX70QW

tsa96 (approved):

tsa96:
> The width for the whole popover has a static width that is relied on by other users of the component. So if I were to put the placeholder button below no-results-found text it looks terrible.
> 
> Now also displays on results found.
> 
> I've tried to minimise the styling changes needed for future use, so there is an extra div wrapper that the credits section uses.

Fair enough, LGTM!

PR #1264, opened 2025-08-08
fix(front): steamID search indefinite wait

### LINKED ISSUE #1260: Searching by steamID with over max limit hangs search
Wherever user-search.component is used you can search by steamID.

If you input a longer than allowed string of numbers, you'll get "Error fetching users!" displayed.
When you change the input afterwards the search just hangs, with indefinite loading spinner.

This is extra annoying in map submission because you have to reload the page, and thus lose all progress with it.

AshenCinders:
Closes #1260 
cheers Gordi!

### Checks

### DIFF
--- a/apps/backend/src/app/validators/is-bigint.validator.ts
+++ b/apps/backend/src/app/validators/is-bigint.validator.ts
@@ -1,5 +1,8 @@
 ﻿import { registerDecorator, ValidationOptions } from 'class-validator';
 
+const MAX_INT64 = BigInt(2 ** 63);
+const MIN_INT64 = -BigInt(2 ** 63);
+
 export function IsBigInt(validationOptions?: ValidationOptions) {
   return function (object: object, propertyName: string) {
     registerDecorator({
@@ -8,7 +11,8 @@ export function IsBigInt(validationOptions?: ValidationOptions) {
       propertyName: propertyName,
       options: validationOptions,
       validator: {
-        validate: (value: unknown) => typeof value === 'bigint',
+        validate: (value: unknown) =>
+          typeof value === 'bigint' && value < MAX_INT64 && value >= MIN_INT64,
         defaultMessage() {
           return `${propertyName} must be a BigInt.`;
         }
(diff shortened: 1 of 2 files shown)


AshenCinders:
I could make the failing sleep test not as precise or someone else needs to rerun tests for me

GordiNoki:
There is a related sentry issue (#6794499518) and the problem is that it tries to get data from db with spcified value as a steamID, and it has type BigInt in db. Even tho nodejs BigInt can support practically any size of a number, postgresql's max BigInt value is 9223372036854775807 (https://www.postgresql.org/docs/current/datatype-numeric.html) which is why it throws on trying to use any bigger numbers. @tsa96 Where should we imlement this range? In the IsBigInt decorator or in the logic itself?

libs/constants/src/consts/limits.const.ts
```diff
@@ -14,6 +14,7 @@ export const MAX_TEST_INVITES = 20;
 export const MIN_PUBLIC_TESTING_DURATION = 7 * 24 * 60 * 60 * 1000;
 export const MAX_CREDITS_EXCEPT_TESTERS = 20;
 export const MAX_BIO_LENGTH = 2000;
+export const STEAM_ID_64_DIGIT_COUNT = 17;
```
tsa96, line 17: Do we really want to code digits like this? Is this something Steam enforce or something? Isn't just limiting to 2^64 okay?

AshenCinders: All the information that I can find says its always 17 digits.
Before (Gordi, I think) does the backend bigint fix, it will hang on any request more than 17 digits.

Currently you get feedback instantly that "it is not a steam64 ID" when not 17 digits, and it doesn't make the extra request to the backend.
But now that there is gonna be a fix on the backend as well, not sure how desirable this fix is?

GordiNoki: We could just implement the same limiting logic on both frontend and backend, so you can just do a max postgres bigint limit there too

tsa96 (approved):

Issue #205, opened 2018-12-05
Implement run styles


Gocnak:
Things like W-only, lowgrav, etc

jason-e:
## Angle Restrictions

Names/terminology used here are placeholders and not necessarily official/final

### Group A (0 degree turn buffer)
- Normal (no key restrictions)
    - Maybe includes backwards as a subtag (need to figure out what the criteria for "backwards" is though)
- SW (W+S)
    - Could include left/right direction only as a subtag
- Surf HSW (WA+SD or WD+SA)
    - Could include backwards as a subtag
    
### Group B (90 degree turn buffer)
- Bhop HSW (WA+WD / SA+SD / WA+SA / WS+SD)
- Other (W+A / W+D / S+A / S+D) 
    - These aren't really a thing but it works the same and would go in this group
    
### Group C (180 degree turn buffer)
- Single key (W / A / S / D)    

## Notes
- Each group is its own leaderboard, and within each leaderboard, each run could be tagged with what specific restrictions that run claimed to have (the only thing that can be verified is which group the run should go in, beyond that the client can fake angles or whatever). You could filter to some specific tag(s) when browsing, but as far as ranking/points goes, there can be no finer breakdown than just these 3 groups.
- Each group is progressively more restrictive, so there could also be one collated leaderboard with all 3 groups and one with B+C as a way to see everything at a glance -- probably not for ranking/points themselves though; not sure how that would work vs the individual groups.
- Technically you could even have 45 deg (e.g. A+WD etc) and 135 deg (e.g. W+WA etc) groups but that would be a bit silly

### Regarding in-game behavior
- The player should be able to choose which of these to run as before they start and the game will apply the necessary restrictions
- Some community servers on other games don't impose angle/key restrictions initially, but determine which mode the player is trying to run based on their inputs. We could have this as a second method but it shouldn't be the only one.
- Basically the player should have the option of having their client explicitly impose restrictions when playing, but the final mode/tags determination would be done independently by the verification backend. Additional tags that the player didn't explicitly select could be added to the run at the time of verification if the run qualifies for them.

jason-e:
## Complete mode hierarchy vs. what we should actually support

![image](https://user-images.githubusercontent.com/2745352/188754360-cc7cda62-4de8-4a6d-9145-48fcb5f2b521.png)

[...]

PR #1268, opened 2025-08-13
Draft: Run styles

Panzerhandschuh:
Closes [#205](https://github.com/momentum-mod/website/issues/205)

I'm opening this up for early review, but we don't need to merge this until after 0.10 releases. TAS, segmented, and easy (low grav, slowmo) run styles still need further discussion.

chrome_OOh1cRpxWS

### Checks

- [ ] __!! DONT IGNORE ME !! I have ran `./create-migration.sh <name>` and committed the migration if I've made DB schema changes__

### DIFF
--- a/apps/backend/src/app/modules/maps/leaderboard-handler.util.ts
+++ b/apps/backend/src/app/modules/maps/leaderboard-handler.util.ts
@@ -7,7 +7,9 @@ import {
   LeaderboardType,
   MapSubmissionSuggestion,
   MapZones,
-  TrackType
+  TrackType,
+  GamemodeStyles,
+  Style
 } from '@momentum/constants';
 import * as Enum from '@momentum/enum';
 import { arrayFrom } from '@momentum/util-fn';
@@ -15,6 +17,7 @@ import { isLinearMainTrack } from '@momentum/formats/zone';
 
 export interface LeaderboardProps
   extends Pick<MapSubmissionSuggestion, 'gamemode' | 'trackType' | 'trackNum'> {
+  style?: Style;
   linear?: boolean;
 }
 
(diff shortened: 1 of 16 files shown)


libs/constants/src/maps/gamemode-styles.map.ts
```diff
@@ -0,0 +1,25 @@
...
+  Gamemode,
+  ReadonlySet<Style>
+> = new Map(
+  [
+    [G.SURF, [S.NORMAL, S.HALF_SIDEWAYS, S.SIDEWAYS]],
+    [G.BHOP, [S.NORMAL, S.HALF_SIDEWAYS, S.SIDEWAYS, S.W_ONLY, S.AD_ONLY]],
+    [G.BHOP_HL1, [S.NORMAL]],
+    [G.CLIMB_MOM, [S.NORMAL, S.NO_TELEPORT]],
+    [G.CLIMB_KZT, [S.NORMAL, S.NO_TELEPORT]],
+    [G.CLIMB_16, [S.NORMAL, S.NO_TELEPORT]],
```
AshenCinders: non TP should be the default, and TP be a style.
But we should discuss this more

AshenCinders: we (I got no negative feedback, at least) want no-TP to be called "Pro" and TP to stay as "TP".
When we add this, we should change stuff in run validator to make sure climb runs are categorized correctly, and added to the correct leaderboards.

apps/backend/src/app/modules/maps/leaderboard-handler.util.ts
```diff
@@ -53,12 +56,20 @@ export function getCompatibleLeaderboards<T extends LeaderboardProps>(
...
+            const validStyles =
+              GamemodeStyles.get(newGamemode) || new Set([Style.NORMAL]);
+
+            return Array.from(validStyles).map((style) => ({
+              trackType,
+              trackNum,
+              linear,
+              gamemode: newGamemode,
+              style
+            }));
```
tsa96: Bit of a nit but you can do `validStyles.map( ... ).toArray()` instead of `Array.from` (generally better practice, similar to doing iterable.Select(...).ToList() over iterable.ToList().Select(...) in LINQ).

Also a very small thing but prefer using `||` over `??` for handling nullish stuff. (Though wouldn't have bother brough it up if not for above stuff)

Panzerhandschuh: `map()` is not a function on `ReadonlySet<T>`. Also I'm already using `||` for handling nullables.

tsa96: The `||` thing I'm not that fussed about especially if you're using it in multiple places, `??` is just more idiomatic in modern JS to `||` for nullish checks.

libs/constants/src/maps/gamemode-styles.map.ts
```diff
@@ -24,3 +24,19 @@ export const GamemodeStyles: ReadonlyMap<
     [G.DEFRAG_VTG, [S.NORMAL]]
   ].map(([k, v]) => [k as Gamemode, new Set((v as [Style | Style[]]).flat())])
 );
+
+export const GamemodeDefaultStyle: ReadonlyMap<Gamemode, Style> = new Map([
```
jason-e: It would be more correct to call this something like "default *UI* style" since it is meant for UI and is not the same thing as the gameplay default. This is particularly important since these types are imported into panorama. Though *really* I don't think it's appropriate for the website to be dictating something like this in panorama since this is not part of some inter-system communication and it creates an unnecessarily tight binding between things (is there another place this could be defined that does not get imported into panorama?)

tsa96 (approved):

PR #1273, opened 2025-08-17
Fix/header buttons

### LINKED ISSUE #1204: Only part of search and bell buttons can be interacted with
In website navbar, only a part of the buttons can be interacted with, which should show a dropdown.

Try clicking around on the button, and you'll see.

Some underlying interactable element needs to be extended.

Image

BJack0:
Closes #1204

Fix header buttons to allow user input and fix m-icon height properties.


AshenCinders (approved):
nice!

PR #1309, opened 2025-09-18
Improve handling of expired access tokens

GordiNoki:
Improves overall experience on dashboard. Since out access tokens are very short lived, and refreshing logic only runs when there is a web request, and not on, for example, switching to another page, users are getting redirected to authentication page much more frequently.

Also fixes a "not found" error when opening a beta map link with expired access token. Since map pages can be accessed without authentication, validity of jwt token is not checked there, so the error returned to client is 403, rather then 401, which triggers refresh.

### Checks


tsa96 (approved):
LGTM, though I'm gonna drastically increase JWT times as well.

PR #1329, opened 2025-10-01
Admin announcement utility

### LINKED ISSUE #1283: Announcement endpoint for admins
Will create announcement notifications to all (?) users.

AshenCinders:
Closes #1283 

Adds creating announcement notifications admin utility.

I plan on converting most forms to use the nonNullableFormBuilder in the near future, so if any of you have opinions on it, please let me know.
Basically it allows not having to write the `{nonNullable: true}` for every formcontrol. By default all controls get set to null on reset (even strings etc!! which is stoopid). `nonNullable` makes it so on reset the form value gets set to the initial value instead.
I am pretty sure you can still directly create nullable formcontrols (nested in nonNullable formgroup) by calling `new Formcontrol` as the value of the field, so changing to it should be really easy.

vivaldi_7rThlUHa2F

### Checks


libs/constants/src/consts/limits.const.ts
```diff
@@ -22,3 +22,4 @@ export const MAX_MAP_NAME_LENGTH = 32; // Seems high but this is actually a cons
 export const MAX_OPEN_MAP_SUBMISSIONS = 200;
 export const MAX_MAP_SUGGESTION_COMMENT_LENGTH = 500;
 export const PRE_SIGNED_URL_EXPIRE_TIME = 5 * 60;
+export const MAX_ADMIN_ANNOUNCEMENT_LENGTH = 500;
```
tsa96, line 25: Wouldn't bother with this in here tbh, very obscure, just hardcode it

PR #1333, opened 2025-10-02
Last parts of map browser filters

### LINKED ISSUE #925: Additional map browser filters
I raced through reworking the map browser pages, and kept the filters fairly minimal. Currently we have four (yes, four :D) browsers, with the following filters:
- Approved Maps
  - Name
  - Gamemode
  - Favorites - either no filter/on/off
  - Completed - no filter/on/off
  - Tiers - double notched slider from 0-10. Only available if you have a gamemode filter
- Map Submissions
  - Name
  - Status - e.g. private testing, public testing
  - Submitter - the user that owns the submitter, not necessarily the author
- Your Maps
  - Name
  - Status - same as Beta Maps, though this *should* include `APPROVED` but won't until I refactor these endpoints a little, coming in next few weeks
- Admin
  - Name
  - Status - any status, including `APPROVED` and `DISABLED`

Here's what I didn't have time to do:
- Sorting (both asc/desc)
  - Date Released [approved, yours, admin]
    - This (desc) should be the default. `MMap` table has an index of `(status, createdAt DESC)` already.
  - Alphabetical
  - Tier
  - Last Played
  - First Created
- Filter by user credit - usually this will be authors (maybe have 4 togglebuttons and have just author on by default) but may as well any credit type. We don't have backend support for this. The `<m-user-select>` component is great for this. We don't need to worry about filtering for multiple users at a time, UI is too annoying.
- Tags - We haven't implemented these yet, I'm going to start working on them in a couple of weeks. Probably best I do this then.

Probably more that we could do, others are welcome to suggest more. Once we have rank points stuff specced out in more detail we could consider possible fancy filtering here.

AshenCinders:
Closes #925 

Adding the remaining sort types are impossible? with the current state of Prisma, and would massively increase complexity of both frontend and backend code for it. So I'm scrapping those last ones.

Adds filter in main map browser for Ranked/Unranked leaderboards.
Adds filter in submission browser for Approving map reviews existing or not.

vivaldi_oBd9WfGDf9
vivaldi_QENJsIlaGQ

### Checks

### DIFF
--- a/apps/frontend/src/app/icons/packs/material-design-icons.icons.ts
+++ b/apps/frontend/src/app/icons/packs/material-design-icons.icons.ts
@@ -27,6 +27,7 @@ export {
   mdiBellAlert,
   mdiBellOutline,
   mdiMagnify,
+  mdiCheckBold,
   mdiCheckCircleOutline,
   mdiPlus,
   mdiPlusThick,
--- a/apps/frontend/src/app/pages/maps/browsers/map-submission-browser.component.html
+++ b/apps/frontend/src/app/pages/maps/browsers/map-submission-browser.component.html
@@ -33,6 +33,16 @@
           appendTo="body"
         />
       </div>
+      <m-n-state-button
+        class="size-8 p-0"
+        [states]="[
+          { icon: 'check-bold', color: 'pale' },
+          { icon: 'check-bold', color: 'green' },
+          { icon: 'check-bold', color: 'red' }
+        ]"
+        mTooltip="Has Approving Reviews"
+        formControlName="hasApprovingReviews"
+      />
       <m-dropdown [entries]="MapSortOptions" [nameFn]="MapSortNameFn" formControlName="sortType" />
     </div>
   </form>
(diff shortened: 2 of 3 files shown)


libs/constants/src/types/queries/map-queries.model.ts
```diff
@@ -54,6 +54,22 @@ type MapsGetAllBaseQuery = {
...
 };
 
+// Keep this in sync with backend MapSortTypeOrder
+export const AllowedMapSortTypes = [
+  MapSortType.DATE_RELEASED_NEWEST,
+  MapSortType.DATE_RELEASED_OLDEST,
+  MapSortType.DATE_CREATED_NEWEST,
+  MapSortType.DATE_CREATED_OLDEST,
+  MapSortType.ALPHABETICAL,
+  MapSortType.REVERSE_ALPHABETICAL,
```
tsa96, line 65: Nit: I'd call this `ALPHABETICAL_REVERSE` for easier searchability

tsa96 (changes requested):

tsa96 (approved):

PR #1374, opened 2025-10-15
feat(back): make "contains" users search work with pagination

GordiNoki:
I hate this solution but nothing else would just work since prisma can't sort with CASE
I don't want to use raw SQL since it would require writing query for all other fields too.
I don't want to use TypedSQL since it doesn't support dynamic WHERE.
I could write a mod to Prisma, but holy... just for this?!

### Checks


GordiNoki:
Will need to update wiki with changes on generating prisma client once this gets in

GordiNoki:
> That would mean any changes to the schema would add massive git diffs as well

Why would it be massive? Generated client files never change if you don't change the schema. And the only way they could change is when you update the prisma version, in which case updating the whole schema is fine

GordiNoki:
> Sorry I must've missed something -- why are we committing the generated Prisma code?

I won't need to generate prisma client on CI then, and then it will work properly without needing a database deploy to generate a client with typed sql. We can just add typed sql files if you don't want the whole prisma client there

PR #1390, opened 2025-10-20
feat(back): use a separate s3 connection for storing runs

GordiNoki:
Storing runs in R2 is not cool, so we'll be moving those to their own thing.

### Checks


tsa96 (changes requested):
Not sure this architecture makes sense, the injection token on one service and not the other is quite ugly. Couldn't we just make an abstract base class, then a `LargeFileStoreService` and a `RunsFileStoreService` handling config stuff in their ctor?

tsa96 (approved):

PR #1399, opened 2025-10-26
feat(discord-internal): add a command to give roles

GordiNoki:


apps/discord-bot-internal/src/interaction-modules/roles.ts
```diff
@@ -0,0 +1,303 @@
...
+            .setName('role')
+            .setDescription('Role')
+            .setRequired(true)
+            .setAutocomplete(true)
+        )
+    )
+    .addSubcommand((builder) =>
+      builder
+        .setName('take')
+        .setDescription('Take a role from a member')
```
tsa96: Not a fan of this naming, wasn't 100% sure what "take" meant at first. I think it'd be better to use "add"/"remove" terminology for this, then something more verbose for the "giveable" stuff, maybe "makeGiveable" / "unmakeGiveable"?

PR #1426, opened 2025-11-20
chore(dev): Introduce Docker development environment

### LINKED ISSUE #1425: Include frontend and backend apps as part of the docker setup
**Is your suggested improvement related to a problem? Please describe.**
Today the app itself is not part of the docker setup.

**Describe the solution you'd like**
The app frontend and backend are now part of the docker up

**Describe alternatives you've considered**
Running the app in the host machine is fine and people should be able to continue to do that. This is just an option for developers.

**Additional context**
None

VitorCaleffi:
Closes #1425

This pull request aims to increase the options of developers to setup their development environment by making the app part of the docker setup.

The ultimate goal is that the developer has a working setup by just cloning the repository and running:
```shell
docker compose up -d
```

This will introduce a new `dev` folder containing the required docker files (and bash script).

What it does:
 - Creates an anonymous volume for the node_modules so it does not conflict with the host node_modules.
 - Mounts the source code in the container so development is a bit more seamless.
 - Runs the migrations on every container start.
 - Runs `npm install` on the container setup.

In order to test this you must modify your `.env` to use the service names instead of `localhost`.

> [!WARNING]  
>For the backend app the app.serve bind is hard coded to `localhost`. This must be changed to support binding to `0.0.0.0` when using docker for app development.

### Testing
1. Follow the env template instructions to fix your local `.env` file.
2. Run `docker compose --profile {full|backend|frontend} up -d`
3. Wait for the containers to be ready and the app running.
4. Access either the frontend or the backend via your normal browser at the expected ports.

### Checks


GordiNoki:
Thanks for a contribution! Sadly, I don't think what you want is feasible. There are multiple issues with how it's implemented right now:
1. Mounting the working directory with a separate anonymous volume for node_modules doesn't work great, since the volume copies over all the files on node_modules, making every start of the container take too much and doubling the drive space you need to run this
2. Networking can be simplified by using network_mode: host, but listen address will still need to be modified, which I'd say is too much of a hustle just for running a dev server
3. There are problems with Prisma client, but I guess adding client generation fixes those.

Also there are problems with your commits, which are all good for a first contribution, but are better to know about if you'll want to contribute more:
1. `bug` is not a type of commit we allow, and `dev` is not a scope we allow. See `commitlint.config.js` for more info about that. This is also caught by our Linting and Formatting CI.
2. All changes to the files added in the PR should be fixed up to the same commit. Judging by "Verified" tag, I assume you made some changes with GitHub editor, which I'd recommend you not to. If you'll need help with figuring out how to do this, you can ask in #website

I will leave a final word to Tom if they think this worth working on more tho.

tsa96 (changes requested):
Okay great, this is looking close to being mergeable. A single commit when merging this is fine, we'll just squash and rebase, don't worry about doing it on your end.

Regarding mine and Gordi's comments, I think you've addressed everything besides 
> > 3. Having to edit .env vars when copying env.TEMPLATE is unfortunate, makes setup more complicated -- which seems to defeat the purpose of having this system in the first place. Previously when we have a containerized setup we have the compose file define an extra env var and used that conditionally in code ([368fadb](https://github.com/momentum-mod/website/commit/368fadb6a53c3039566966f77c01b7aa3ed6911a) for part of it, can't find the rest). Again, if the purpose of this system is to let new devs get up and running with a single command, it seems best to avoid requiring altering the .env template as much as possible.
> 
> We could also introduce a new env template like `env.DOCKER` if thats fine for you.

Okay yeah, let's go with that and I'll update the setup guide when merging this. The old conditional `IS_DOCKERIZED_MODE` stuff I was doing in code would be too ugly.

PR #1511, opened 2026-07-06
Clustered WebSockets + Run session refactor

Panzerhandschuh:
### Checks


apps/backend/src/app/modules/game-connection/game-connection.gateway.ts
```diff
@@ -0,0 +1,219 @@
...
+@WebSocketGateway({ path: '/game' })
+export class GameConnectionGateway {
+  private readonly logger = new Logger(GameConnectionGateway.name);
+
+  constructor(
+    private readonly valkey: ValkeyService,
+    @Inject(EXTENDED_PRISMA_SERVICE) private readonly db: ExtendedPrismaService
+  ) {}
+
+  @SubscribeMessage('session.create')
```
tsa96: "session" is quite generic, considering we'll probably be having map sessions as well. Wouldn't "runsession" be better?

apps/backend/src/app/modules/game-connection/game-connection.gateway.ts
```diff
@@ -0,0 +1,219 @@
...
+        id,
+        userID,
+        createdAt: createdAtDate,
+        ...leaderboardData,
+        timestamps: [
+          { majorNum: 1, minorNum: 1, time: 0, createdAt: createdAtDate }
+        ]
+      }
+    };
+  }
```
tsa96: The bulk of these methods should be in a separate .service.ts file. Gateway files are effectively the same as controllers, they define the method names and DTOs for body and return time, DB/Valkey etc should be service logic.

apps/backend/src/app/modules/game-connection/game-connection.gateway.ts
```diff
@@ -0,0 +1,219 @@
...
+  @SubscribeMessage('session.create')
+  async createSession(
+    @ConnectedSocket() client: AuthenticatedWebSocket,
+    @MessageBody()
+    data: {
+      mapID: number;
+      gamemode: number;
+      trackType: number;
+      trackNum: number;
+    }
```
tsa96: If we have class-validator setup for ws gateways, these can be annotated with DTOs like controllers do, and that'll perform validation. Otherwise people could set any garbage data and make us error.

apps/backend/src/app/modules/game-connection/game-connection.gateway.ts
```diff
@@ -0,0 +1,219 @@
...
+  ): Promise<WsResponse<unknown>> {
```
tsa96: This should be typed, ideally with a DTO, like controllers do.

tsa96: Actually okay, from discussion we're agreeing to avoid returning DTOs and just use JSON.stringify, but let's still write an explicit type for these.

package.json
```diff
@@ -37,6 +37,7 @@
     "@nestjs/schedule": "6.1.2",
     "@nestjs/swagger": "11.2.0",
     "@nestjs/throttler": "6.5.0",
+    "@nestjs/websockets": "^11.1.24",
```
tsa96: Please pin all deps (no ^)

tsa96 (changes requested):

apps/backend/src/app/modules/session/run/run-session.service.ts
```diff
@@ -39,15 +51,211 @@ import { LeaderboardRunsDbService } from '../../runs/leaderboard-runs-db.service
...
-    private readonly leaderboardRunsDbService: LeaderboardRunsDbService
+    private readonly leaderboardRunsDbService: LeaderboardRunsDbService,
+    private readonly killswitch: KillswitchService
   ) {}
 
+  //#region Run Session (game WebSocket)
+
+  // These handle the live `runsession.*` messages from the game client. They
+  // return the response payload (a success DTO, an error DTO, or null for a
+  // bare ack); the GameConnectionGateway frames them into `{ event, data }`.
```
tsa96: This just seems like LLM-babble to me. The message name isn't at all relevant to the service, that's for the controller. Then you're basically just summarising what the types already tell you.

Often when you ask them to some a specific change it'll insert comments like these sort of justifying the change, and it really isn't helpful.

apps/backend/src/app/modules/session/run/run-session.service.ts
```diff
@@ -39,15 +51,211 @@ import { LeaderboardRunsDbService } from '../../runs/leaderboard-runs-db.service
...
+
+  /**
+   * `runsession.create`: start a new session for the user on a leaderboard,
+   * clearing any existing session for the same track first.
+   */
```
tsa96, line 69: As above, message name isn't relevant here. I guess the rest of the comment is okay. I know I'm nitpicking here, but I'm wary of filling the repo with LLM-generated stuff.

apps/backend/src/app/modules/game-connection/game-connection.gateway.ts
```diff
@@ -26,14 +33,8 @@ export class GameConnectionGateway {
...
-    @MessageBody()
-    data: {
-      mapID: number;
-      gamemode: number;
-      trackType: number;
-      trackNum: number;
-    }
-  ): Promise<WsResponse<unknown>> {
+    @MessageBody() data: CreateRunSessionDto
+  ): Promise<WsResponse<RunSessionResponseDto | RunSessionErrorDto>> {
```
tsa96: When we spoke about this in voice the other day I agreed that we could skip using DTOs for responses here and just use plain types. But, if we're going to use DTOs, create a wrapper DTO like this:

```ts
export class WsResponseDto<T extends Record<string, unknown>> {
  @IsString()
  readonly event: string;

  readonly data: T | { error: string };

  constructor(c: { new (): T }, event: string, data: T | { error: string }) {
    this.event = event;
    this.data = 'error' in data ? data : DtoFactory(c, data);
  }
}
```

That way class-transformer will apply any transformations to `T`, also it actually erases types so you don't need to call `.toISOString` explicitly (`Date`s stringify to ISO strings anyway).

Writing this out made me realise that we should probably be using Nest's exception filters (https://docs.nestjs.com/websockets/exception-filters) since they let you throw an unhandled error to return an error response (esp helpful for e.g. unhandled Prisma errors), and sending stuff to Sentry. But let's leave that for now, I'll probably implement in the future.

tsa96 (changes requested):

apps/backend/src/app/modules/session/run/run-session.service.ts
```diff
@@ -46,138 +49,203 @@ import { LeaderboardRunsDbService } from '../../runs/leaderboard-runs-db.service
...
+  /**
+   * Start a new session for the user on a leaderboard, clearing any existing
+   * session for the same track first.
+   */
   async createSession(
     userID: number,
-    body: CreateRunSessionDto
-  ): Promise<RunSessionDto> {
+    data: CreateRunSessionDto
+  ): Promise<(RunSession & { id: number }) | { error: string }> {
```
tsa96, line 73: Usually I do the DTO transform in the service just so we don't need clunky types like this, but not really a big deal. If I move us to using exception handling stuff we can simplify this then.

tsa96:
Only remaining thing is commitlint complaining about `feat(backend)` instead of `feat(back)`

tsa96 (approved):

PR #1513, opened 2026-08-10
Player Reporting System

Panzerhandschuh:
Report Queue:
chrome_1TCXXkiWSb

Update Report Dialog:
image

Admin Activity:
image

Player Profile:
image

Admin Settings:
image

### Checks


libs/db/src/schema.prisma
```diff
@@ -91,12 +93,40 @@ model Report {
   resolver   User? @relation("report_resolver", fields: [resolverID], references: [id], onDelete: SetNull)
   resolverID Int?
 
+  /// Chat/voice bans issued while resolving this report, if any
+  chatBans ChatBan[]
```
tsa96: None of the other relations here have comments like this, and I don't think they're necessary. If you're going to use LLMs for this, please try to make the output match the existing file.

libs/db/src/schema.prisma
```diff
@@ -91,12 +93,40 @@ model Report {
...
+  expiresAt DateTime? /// If NULL, the ban is permanent
+  reason    String?
+
+  target   User @relation("chatban_target", fields: [targetID], references: [id], onDelete: Cascade)
+  targetID Int
+
+  issuer   User? @relation("chatban_issuer", fields: [issuerID], references: [id], onDelete: SetNull)
+  issuerID Int?
+
+  /// The report whose resolution created this ban, if any
```
tsa96: Again, this comment feels unnecessary.

apps/backend/src/app/dto/chat-ban/chat-ban-input.dto.ts
```diff
@@ -0,0 +1,49 @@
...
+  DateString,
+  UpdateChatBan
+} from '@momentum/constants';
+import { ApiPropertyOptional } from '@nestjs/swagger';
+import { IsDateString, IsOptional, IsString } from 'class-validator';
+import { EnumProperty } from '../decorators';
+
+// Input DTOs for chat/voice bans. Kept in their own leaf module (no other DTO
+// imports) so `report.dto` can pull in `CreateChatBanDto` without forming an
+// import cycle with the full `ChatBanDto` (which references `UserDto`).
```
tsa96: I think the `lazy: true` option in `NestedDto` should resolve any circular dependency issues. And what you're doing this is pervasive throughout the codebase -- the fact an LLM is clearly doing different from everywhere else should be ringing alarm bells, it's your responsibility not to let it hack together its own solutions. If you can't figure it out, better to ask me during dev IMO.

tsa96: Once you resolve this you can use `PickType` on `ChatBanDto`, like other DTOs do.

apps/backend/src/app/dto/user/user.dto.ts
```diff
@@ -22,12 +25,40 @@ import {
...
+  NestedProperty
+} from '../decorators';
 import { IsCountryCode, IsSteamCommunityID } from '../../validators';
 import { ProfileDto } from './profile.dto';
 import { UserStatsDto } from './user-stats.dto';
 import { UpdateSocialsDto } from './socials.dto';
 
+// Minimal representation of an active chat/voice ban, embedded in the local
+// user's GET /user response. Defined here (rather than importing the full
+// ChatBanDto) so UserDto stays free of the ChatBan->User import cycle.
```
tsa96: More import cycle system weirdness again

apps/backend/src/app/modules/admin/admin.service.ts
```diff
@@ -413,6 +417,18 @@ export class AdminService {
 
     if (!report) throw new NotFoundException('Report not found');
 
+    const issuingBans = reportDto.bans && reportDto.bans.length > 0;
```
tsa96: bans?.length > 0

apps/backend/src/app/modules/admin/admin.service.ts
```diff
@@ -438,6 +454,94 @@ export class AdminService {
...
+    dto: UpdateChatBanDto
+  ): Promise<void> {
+    const ban = await this.db.chatBan.findUnique({ where: { id: banID } });
+
+    if (!ban) throw new NotFoundException('Ban not found');
+
+    const data: Prisma.ChatBanUpdateInput = {};
+    if (dto.expiresAt !== undefined)
+      data.expiresAt = dto.expiresAt ? new Date(dto.expiresAt) : null;
+    if (dto.reason !== undefined) data.reason = dto.reason;
```
tsa96: I think Prisma does the ISO string to DateTime automatically, we don't use this for other submitted dates. Is the `new Date` call actually needed?
So I think you can do
```
const data: Prisma.ChatBanUpdateInput = {
  expiresAt: dto.expiresAt ?? null,
  reason: dto.reason ?? null
};
```

apps/backend/src/app/modules/users/users.service.ts
```diff
@@ -96,9 +96,25 @@ export class UsersService {
...
+    // reads these to block communication and show a toast.
+    if (includeChatBans) {
+      include.chatBans = {
+        where: {
+          OR: [{ expiresAt: null }, { expiresAt: { gt: new Date() } }]
+        },
+        select: { type: true, expiresAt: true, reason: true }
+      };
+    }
+
```
tsa96, line 118: I'd call this param `localUser` not `includeChatBans`, we want to express that the method's functionality varies according to the endpoint. Also this comment is a bit confusing, of course users will be able to know each other's bans, we're just not exposing the reason or duration.

apps/backend/src/app/dto/report/report.dto.ts
```diff
@@ -78,7 +79,16 @@ export class CreateReportDto extends PickType(ReportDto, [
...
+  implements UpdateReport
+{
+  @NestedProperty(CreateChatBanDto, {
+    isArray: true,
+    required: false,
+    description:
+      'Chat/voice bans to issue against the reported user when resolving a ' +
+      'player report. Only honoured when the report is being resolved.'
+  })
+  readonly bans?: CreateChatBanDto[];
```
tsa96: Clearer to call this `chatBans`, surely. Maybe other bans (e.g. cheating) could use the same DTO in the future, but then we'd want to rename `ChatBan` to something more generic like `Ban`.

apps/backend-e2e/src/chat-bans.e2e-spec.ts
tsa96, line 1: Should be in admin.e2e-spec.ts as with other stuff on this endpoint. I know that file is massive, but each separate E2E test file spins up a new Nest instance and Prisma connection which slows down CI quite a bit. I'd like to break up E2E stuff more in the future (maybe a good task for LLMs to work on) but either needs very different Jest configuration or moving to Vitest/node:test.

apps/frontend/src/app/pages/profile/profile-chat-bans/profile-chat-bans.component.html
```diff
@@ -0,0 +1,84 @@
...
+      <p class="text-gray-300">This user has no communication bans.</p>
+    }
+
+    <div class="flex flex-col gap-3">
+      @for (ban of bans; track $index) {
+        <div class="card-inset flex flex-col gap-1 p-3">
+          <div class="flex items-center gap-2">
+            <span class="font-medium">{{ ChatBanTypeNames.get(ban.type) }}</span>
+            @if (isActive(ban)) {
+              <span class="rounded bg-red-800 px-2 py-0.5 text-xs">Active &middot; {{ expiryLabel(ban) }}</span>
```
tsa96: I think you mentioned, we should be able to see exact date of this. `m-tooltip` just containing date.toUTCString should be fine

tsa96: Also, `&middot;` escape sequence like this? Feeling too much like LLM frontend slopt to me lol, use a hyphen.

apps/frontend/src/app/pages/profile/profile-chat-bans/profile-chat-bans.component.html
```diff
@@ -0,0 +1,84 @@
...
+          <p class="text-sm text-gray-300">
+            @if (ban.createdAt) {
+              <span>Issued {{ ban.createdAt | date: 'medium' }}</span>
+            }
+            @if (ban.issuer) {
+              <span>
+                by <a class="hover:!text-blue-500" [routerLink]="'/profile/' + ban.issuer.id">{{ ban.issuer.alias }}</a></span
+              >
+            }
+            <span> &middot; {{ ban.expiresAt ? 'Expires ' + (ban.expiresAt | date: 'medium') : 'Never expires' }}</span>
```
tsa96: `&middot;` again, yucky

libs/constants/src/types/queries/report-queries.model.ts
```diff
@@ -9,10 +9,14 @@ export type ReportGetQuery = PagedQuery & {
...
-  'data' | 'type' | 'category' | 'message'
->;
+export type CreateReport = Pick<Report, 'type' | 'category' | 'message'> &
+  Partial<Pick<Report, 'data'>> & {
+    // Provide EITHER `data` (the reported object's ID, which the website already
+    // knows) or `targetSteamID` (the SteamID of a reported player, which is all
+    // the game knows — the backend resolves it to a user ID). `targetSteamID` is
+    // only valid for PLAYER_REPORTs.
+    targetSteamID?: string;
+  };
```
tsa96, line 15: Better to document on the DTO so works in Swagger than comments like this, generally don't use this fine for documentation.

apps/frontend/src/app/pages/profile/profile-chat-bans/profile-chat-bans.component.html
```diff
@@ -1,7 +1,10 @@
 @if (activeBans.length > 0 || (canManage && bans.length > 0)) {
   <div class="flex flex-wrap items-center gap-2">
     @for (ban of activeBans; track $index) {
-      <span class="rounded bg-red-800 px-2 py-1 text-sm font-medium">
+      <span
+        class="rounded bg-red-800 px-2 py-1 text-sm font-medium"
+        [title]="ban.expiresAt ? 'Expires ' + (ban.expiresAt | date: 'medium') : 'This ban does not expire'"
+      >
         {{ ChatBanTypeNames.get(ban.type) }} ban &middot; {{ expiryLabel(ban) }}
```
tsa96: Sorry but these `&middot;` escape sequences... Feeling too much like LLM frontend slopt to me lol, use a hyphen

tsa96 (changes requested):
Haven't much time look at frontend side, logic looks fine don't really care about the styling for admin stuff.

Can you add reports + chatban integration to the seed script as well please? Want to be able to look at the frontend myself with some mock data. Seed script isn't that important, LLM should handle it easily.

tsa96:
From screenshots the report and resolution message textboxes look very small, just adding a scrollbar feels like a bit of a cop-out. Probably best to make each report item collapsed and expandable, could also my [m-accordion component](https://github.com/momentum-mod/website/blob/main/apps/frontend/src/app/components/accordion/accordion.component.ts) for that.

Panzerhandschuh:
> From screenshots the report and resolution message textboxes look very small, just adding a scrollbar feels like a bit of a cop-out. Probably best to make each report item collapsed and expandable, could also my [m-accordion component](https://github.com/momentum-mod/website/blob/main/apps/frontend/src/app/components/accordion/accordion.component.ts) for that.

I'm not a fan of how the accordion component looks when applied to the report queue items. I'll probably leave this as-is unless you have a better suggestion.

tsa96:
> > From screenshots the report and resolution message textboxes look very small, just adding a scrollbar feels like a bit of a cop-out. Probably best to make each report item collapsed and expandable, could also my [m-accordion component](https://github.com/momentum-mod/website/blob/main/apps/frontend/src/app/components/accordion/accordion.component.ts) for that.
> 
> I'm not a fan of how the accordion component looks when applied to the report queue items. I'll probably leave this as-is unless you have a better suggestion.

Can't you restyle it in some way then? Or add a small togglebutton absolute positioned in top right of the text boxes, that makes the box take up the full height of the text. You could also shrink the width of other columns, especially "Type" -- change to just "Map"/"Player" and drastically shrink the width. Trying to read lengthy text in that box sounds very aggravating otherwise. But also, since we're going to show player's message contents there in the future, we're definitely going to need more space eventually. I would just do the accordion now, rather than inevitably changing to that/something similar later when the report content becomes too large to condense into a single line like this.

Panzerhandschuh:
> > > From screenshots the report and resolution message textboxes look very small, just adding a scrollbar feels like a bit of a cop-out. Probably best to make each report item collapsed and expandable, could also my [m-accordion component](https://github.com/momentum-mod/website/blob/main/apps/frontend/src/app/components/accordion/accordion.component.ts) for that.
> > 
> > 
> > I'm not a fan of how the accordion component looks when applied to the report queue items. I'll probably leave this as-is unless you have a better suggestion.
> 
> Can't you restyle it in some way then? Or add a small togglebutton absolute positioned in top right of the text boxes, that makes the box take up the full height of the text. You could also shrink the width of other columns, especially "Type" -- change to just "Map"/"Player" and drastically shrink the width. Trying to read lengthy text in that box sounds very aggravating otherwise. But also, since we're going to show player's message contents there in the future, we're definitely going to need more space eventually. I would just do the accordion now, rather than inevitably changing to that/something similar later when the report content becomes too large to condense into a single line like this.

Lmk if this looks good:
chrome_fWAPbrsXku

tsa96:
> > > > From screenshots the report and resolution message textboxes look very small, just adding a scrollbar feels like a bit of a cop-out. Probably best to make each report item collapsed and expandable, could also my [m-accordion component](https://github.com/momentum-mod/website/blob/main/apps/frontend/src/app/components/accordion/accordion.component.ts) for that.
> > > 
> > > 
> > > I'm not a fan of how the accordion component looks when applied to the report queue items. I'll probably leave this as-is unless you have a better suggestion.
> > 
> > Can't you restyle it in some way then? Or add a small togglebutton absolute positioned in top right of the text boxes, that makes the box take up the full height of the text. You could also shrink the width of other columns, especially "Type" -- change to just "Map"/"Player" and drastically shrink the width. Trying to read lengthy text in that box sounds very aggravating otherwise. But also, since we're going to show player's message contents there in the future, we're definitely going to need more space eventually. I would just do the accordion now, rather than inevitably changing to that/something similar later when the report content becomes too large to condense into a single line like this.
> 
> Lmk if this looks good:
> chrome_fWAPbrsXku

Nice, good enough for admin stuff. Will rereview everything soon

tsa96 (approved):

